PR #19, opened 2015-01-26
setup.py: Make URL on PyPI point to GitHub

msabramo:
rather than pointing at itself, which isn't very useful.


mgedmin:
:sparkles: Thank you! :sparkles:

PR #20, opened 2015-01-26
docs/README.rst: Small tweaks

msabramo:


docs/README.rst
```diff
@@ -34,11 +34,11 @@ Interfaces are defined using Python class statements:
   ...    def bar(q, r=None):
   ...        """bar blah blah"""
 
-In the example above, we've created an interface, `IFoo`.  We
-subclassed `zope.interface.Interface`, which is an ancestor interface for
+In the example above, we've created an interface, :py:data:`IFoo`.  We
```
tseaver: `:py:data:IFoo` does not seem correct:

> .. py:data:: name
>
>     Describes global data in a module, including both variables and values used as “defined constants.” Class and object attributes are not documented using this environment.

An interface is much more "class-like" than "data-like".  Also `:py:` is the default domain (just as for `:class:`).

msabramo: Thanks, updated.

PR #29, opened 2015-11-24
zope.interface does not in fact depend on setuptools at install time.

Julian:
At most, it depends on it to actually run the setup.py (but this appears to also not be the case).

Putting install_requires=["setuptools"] does bad things to pip install
-U. pip doesn't have a minimal upgrade mode currently, so having this there means users installing zope will have their setuptools updated
when they upgrade zope.interface, which is a risky upgrade and one that
is not necessary for zope.interface to run.

This especially breaks things when setuptools has an accidentally bad release (as it currently appears), because zope will continuously try to upgrade to it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,6 @@ else:
         include_package_data = True,
         zip_safe = False,
         tests_require = tests_require,
-        install_requires = ['setuptools'],
         extras_require={'docs': ['Sphinx', 'repoze.sphinx.autointerface'],
                         'test': tests_require,
                         'testing': testing_extras,


Julian:
Apologies if I'm totally wrong by the way and there is somewhere in z.i that is in fact doing `import setuptools` as part of running it -- I did a simple grep, and based on what I know z.i does, I highly doubt it actually does, but if I missed it and this is intentionally there, please lemme know that I'm way off.

tseaver:
@Julian the irony here is that distribute/setuptools actually bitched for a long time about the `namespace_packages=['zope']` declaration if we *didn't* declare `install_requires=['setuptools']`.  We are able to [fall back to `pkgutil`](https://github.com/zopefoundation/zope.interface/blob/b2589c02b0208b8db25cbe39844885ba611f2bc3/src/zope/__init__.py) if `setuptools` (which provides the `pkg_resources` package) is missing.

Julian:
Interesting. Sorry, I forgot to grep for `pkg_resources`.

I do remember asking once if there was a reason to ever use `pkg_resources` over `pkgutil`.

But if that's the case (that you look for `pkg_resources` then this is probably not completely trivial to remove without considering the above ^).

PR #43, opened 2016-08-02
Remove the _u compat function.

jamadden:
Since we no longer support 3.2 we can use the literal syntax.

This was done with a quick sed script.


jamadden:
Travis is (slowly) going green---those OS X builds really drag things down.

jamadden:
Yay, green!

tseaver:
LGTM

PR #44, opened 2016-08-03
Make declarations.Implements sortable.

### LINKED ISSUE #42: Make declarations.Implements (and possibly Provides) comparable?
Recently we had the need to do some `IComponents`-like things on some very large persistent data sets. So we started with a subclass of `AdapterRegistry` and made it use BTrees instead of dicts. Everything works fine so long as you stay in the realm of interfaces, but as soon as you try to do something based on classes, like `components.registerAdapter(factory, provided=IFoo, required=(str,))`  you blow up with `TypeError: Object has default comparison`. 

Actually it's slightly worse than that. Some of the dicts used in the implementation of Components (which we also replaced with BTrees) use keys that are tuples. Tuples are automatically comparable, even if their elements aren't. But having non-comparable keys (or rather, keys that change across processes) is a sure way to corrupt the BTree.

I looked into it, and was pretty surprised to find that not only are `Implements` and `Provides` not comparable, they actually rely on the default, identity-based, `__hash__` and `__eq__` inherited from object. Everything pretty much happens to work out for a regular `PersistentAdapterRegistry` because of the way these objects are pickled (in Implement's case, via `__reduce__` and a call to `implementedBy`, which aggressively caches the same object, so identities stay the same in a single process even across pickling).

Is it a bad idea to make `Implements` comparable so it can be used in BTrees? To work for the `IComponents` use-case, it would have to be mutually-comparable with `InterfaceClass`, but it already has `__name__` and `__module__` (which is what `InterfaceClass` uses), so that might just work out nearly automatically.

If there's consensus this is a reasonable idea, I can work on a PR. (In our particular use-case, we were able to simply prohibit class-based registrations as they don't occur in our data set.)

jamadden:
Fixes #42

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changes
 4.2.1 (unreleased)
 ------------------
 
-- TBD
+- Add the ability to sort the objects returned by ``implementedBy``.
+  This is compatible with the way interface classes sort so they can
+  be used together in ordered containers like BTrees.
+  (https://github.com/zopefoundation/zope.interface/issues/42)
 
 4.2.0 (2016-06-10)
 ------------------
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -30,7 +30,6 @@ import sys
 from types import FunctionType
 from types import MethodType
 from types import ModuleType
-import warnings
 import weakref
 
 from zope.interface.advice import addClassAdvisor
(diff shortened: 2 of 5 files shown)


jamadden:
I also ran the `zope.component` test suite against this branch and didn't run into any issues locally.

The `__eq__` and `__hash__` behaviour should be exactly the same as it was, only sorting is new. So far as I see, even though `__eq__` and `__hash__` aren't technically fully consistent with sorting, practically speaking they should be, given the caching that goes on.

jamadden:
I only did this for `Implements` because as far as I can tell `Components` only ever stores `implementedBy` values in its registrations. It might request `providedBy`, but only ever uses the first element (an interface) of those.

mgedmin:
LGTM, after a brief look.

PR #45, opened 2016-08-04
Make setuptools a hard dependency of setup.py

### LINKED ISSUE #13: W/o setuptools, "setup.py install" skips zope/interface/common/
In https://bugs.launchpad.net/zope.interface/+bug/611362, @exarkun reported:

> After installing the 3.6.1 release from PyPI, the zope.interface.common package is not available:

    exarkun@boson:~/Downloads$ tar xzf zope.interface-3.6.1.tar.gz
    exarkun@boson:~/Downloads$ cd zope.interface-3.6.1/
    exarkun@boson:~/Downloads/zope.interface-3.6.1$ python setup.py install --quiet --prefix /tmp/ZOPE-TEST
    /usr/lib/python2.6/distutils/dist.py:266: UserWarning: Unknown distribution option: 'test_suite'
    ....
    exarkun@boson:~/Downloads/zope.interface-3.6.1$ PYTHONPATH=/tmp/ZOPE-TEST/ python -c 'from zope.interface import common'
        File "<string>", line 1, in <module>
    ImportError: cannot import name common
    exarkun@boson:~/Downloads/zope.interface-3.6.1$

and supplied a patch:

  https://bugs.launchpad.net/zope.interface/+bug/611362/+attachment/1471651/+files/setup.patch

I followed up:

> Hmm, this seems to be true only when installing without setuptools:

@glyph followed up:

> I just heard about a case of this happening on Ubuntu 12.10. I suspect it's a bug in setuptools or distutils or ubuntu's patching of same, not zope.interface; can someone comment on what mechanism ZI uses to include the 'common' package? I don't see it in setup.cfg, or a manifest file... I am guessing there's some setuptools magic which is supposed to make it get included.
> Also of note; setuptools *was* installed on the system where this happened. It was zope.interface 4.0.3 in this case.

### LINKED ISSUE #14: Fix setup.py when using plain distutils
In https://bugs.launchpad.net/zope.interface/+bug/629086, @zseil reported:

> Support for installing zope.interface with plain distutils has deteriorated a bit over the recent releases. There are currently two problems with the setup script - the list of packages is incomplete (reported as bug #611362) and the coptimization extension never gets built when not using setuptools. The attached patch fixes both problems for me. It also removes Python 2.3 support from the setup script, since zope.interface itself doesn't support that version anymore.
>
> (Tested on Windows, Python 2.5 with a C compiler and Python 2.6 without a C compiler, in both cases without setuptools installed)

with patch:  https://bugs.launchpad.net/zope.interface/+bug/629086/+attachment/1537986/+files/distutils-compatibility.patch

jamadden:
Uses `find_packages` instead of a manual list.

Fixes #13.
Fixes #14.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@ Changes
   be used together in ordered containers like BTrees.
   (https://github.com/zopefoundation/zope.interface/issues/42)
 
+- Make ``setuptools`` a hard dependency of ``setup.py``.
+  (https://github.com/zopefoundation/zope.interface/issues/13)
+
 4.2.0 (2016-06-10)
 ------------------
 
--- a/setup.py
+++ b/setup.py
@@ -23,14 +23,14 @@ import os
 import platform
 import sys
 
+from setuptools import setup, Extension, Feature
+from setuptools.command.build_ext import build_ext
+from setuptools import find_packages
+
 from distutils.errors import CCompilerError
 from distutils.errors import DistutilsExecError
 from distutils.errors import DistutilsPlatformError
 
-try:
-    from setuptools.command.build_ext import build_ext
-except ImportError:
-    from distutils.command.build_ext import build_ext
 
 class optional_build_ext(build_ext):
     """This class subclasses build_ext and allows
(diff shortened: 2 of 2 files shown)


tseaver:
LGTM

PR #48, opened 2016-08-19
Use dictionary lookups for testing subscribed status.

### LINKED ISSUE #46: Components.register/unregisterUtility is slow (linear with size of all registrations)
We've noticed that as a component registry gets larger, the time taken to add or remove utilities increases with the total number of all utility registrations. It doesn't take very many utilities to make registering and unregistering a painfully slow process.

Here's some pseudo-code that demonstrates, using IPython's `%time` magic to record how long it takes to register 1000 new utilities:
```python
In [1]: import zope.interface.registry
In [2]: import zope.interface
In [3]: class IFoo(zope.interface.Interface): pass
In [5]: @zope.interface.implementer(IFoo)
   ...: class Foo(object): pass
   ...:
In [6]: reg = zope.interface.registry.Registry()
In [16]: def add_to_reg(reg, count):
    ...:     for _ in range(count):
    ...:         comp = Foo()
    ...:         reg.registerUtility(component=comp, name=str(id(comp)))
In [12]: %time add_to_reg(reg, 1000)
CPU times: user 785 ms, sys: 7.23 ms, total: 792 ms
Wall time: 810 ms

In [17]: %time add_to_reg(reg, 1000)
CPU times: user 2.32 s, sys: 19.1 ms, total: 2.34 s
Wall time: 2.4 s

In [18]: %time add_to_reg(reg, 1000)
CPU times: user 3.92 s, sys: 28.8 ms, total: 3.95 s
Wall time: 4.05 s

In [19]: %time add_to_reg(reg, 1000)
CPU times: user 5.57 s, sys: 41.1 ms, total: 5.61 s
Wall time: 5.76 s

In [20]: %time add_to_reg(reg, 1000)
CPU times: user 7.29 s, sys: 55.7 ms, total: 7.34 s
Wall time: 7.53 s
```

In other words, by the time there are 5,000 objects in the registry, adding a new utility takes 72ms. And it just keeps growing.

The culprit is in checking whether or not the component is already subscribed. This is a linear process that looks at every entry in the `_utility_registrations` dict (likewise for unregister):
```python
def registerUtility(...):
         ...
        subscribed = False
        for ((p, _), data) in iter(self._utility_registrations.items()):
            if p == provided and data[0] == component:
                subscribed = True
                break

        self._utility_registrations[(provided, name)] = component, info, factory
        self.utilities.register((), provided, name, component)

        if not subscribed:
            self.utilities.subscribe((), provided, component)
```

If we could safely assume that all the `components` are hashable, we could substantially optimize this by using an auxilliary dictionary (keyed on *just* `provided`) and a `set` of components. Something like this:
```python
    @property
    def _v_utility_opt(self):
        if '_v_utility_opt' in self.__dict__:
            return self.__dict__['_v_utility_opt']
        from collections import defaultdict
        opt = defaultdict(set)
        for ((p, _), data) in iter(self._utility_registrations.items()):
            opt[p].add(data[0])
        self.__dict__['_v_utility_opt'] = opt
        return opt

    def _subscribed(self, provided, component):
        return component in self._v_utility_opt[provided]

    def registerUtility(self, component=None, provided=None, name=_u(''),
                        info=_u(''), event=True, factory=None):
        ...
        subscribed = self._subscribed(provided, component)

        self._utility_registrations[(provided, name)] = component, info, factory
        self._v_utility_opt[provided].add(component)
        self.utilities.register((), provided, name, component)

        if not subscribed:
            self.utilities.subscribe((), provided, component)
```

That makes adding and removing O(1) and nearly independent of the size of the components registered:
```python
In [25]: reg = zope.interface.registry.Components()

In [26]: %time add_to_reg(reg, 1000)
CPU times: user 53.7 ms, sys: 1.49 ms, total: 55.2 ms
Wall time: 54.6 ms

In [27]: %time add_to_reg(reg, 1000)
CPU times: user 57.5 ms, sys: 1.57 ms, total: 59.1 ms
Wall time: 58.8 ms

In [28]: %time add_to_reg(reg, 1000)
CPU times: user 79.5 ms, sys: 1.84 ms, total: 81.4 ms
Wall time: 81.1 ms

In [29]: %time add_to_reg(reg, 1000)
CPU times: user 72.8 ms, sys: 1.79 ms, total: 74.6 ms
Wall time: 74.2 ms

In [30]: %time add_to_reg(reg, 1000)
CPU times: user 73 ms, sys: 1.12 ms, total: 74.1 ms
Wall time: 74.2 ms

In [31]: %time add_to_reg(reg, 1000)
CPU times: user 115 ms, sys: 5.19 ms, total: 120 ms
Wall time: 120 ms

In [32]: len(reg._utility_registrations)
Out[32]: 6000
```

(Much better. And imagine the effect it could have on startup time and testing time for large apps, independent of registering utilities at runtime.)

Is this something we can assume? Or is there some other way to make this process faster? Or am I missing something altogether?

If there's a consensus here, I'm happy to write the code (but it won't be for a week or so).

jamadden:
Fixes #46.

Benchmarks show a dramatic improvement; not quite as good as the demonstration in #46 because the implementation had to be a bit more complex to properly handle unregistration, but still very good. Here it is with 20,000 items already registered:
```
   %time add_to_reg(reg, 1000)
   CPU times: user 190 ms, sys: 19.3 ms, total: 209 ms
   Wall time: 203 ms
```

Here's a profile with about 100,000 utilities registered:
```
 %prun add_to_reg(reg, 1000)
         80005 function calls (79005 primitive calls) in 0.713 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1000    0.596    0.001    0.621    0.001 adapter.py:202(subscribe)
    12000    0.016    0.000    0.019    0.000 interface.py:518(__hash__)
     1000    0.010    0.000    0.709    0.001 registry.py:206(registerUtility)
3000/2000    0.009    0.000    0.014    0.000 interface.py:255(interfaces)
     2000    0.008    0.000    0.022    0.000 adapter.py:637(changed)
     1000    0.008    0.000    0.025    0.000 registry.py:498(_getUtilityProvided)
     9000    0.008    0.000    0.017    0.000 {method 'get' of 'dict' objects}
     1000    0.008    0.000    0.027    0.000 adapter.py:102(register)
     2000    0.006    0.000    0.008    0.000 adapter.py:450(changed)
     1000    0.005    0.000    0.663    0.001 registry.py:145(registerUtility)
```

I was very careful not to change the pickle at all. (Thus the weakrefs.)

zope.interface and zope.component tests have been run and both pass. (It was necessary to account for the underlying objects changing because of the way zope.component cleans up after tests.) I show 100% test coverage still.

### DIFF
--- a/src/zope/interface/registry.py
+++ b/src/zope/interface/registry.py
@@ -41,17 +41,16 @@ from zope.interface._compat import CLASS_TYPES
 from zope.interface._compat import STRING_TYPES
 
 
-class _CacheList(object):
+class _UnhashableComponentCounter(object):
     # defaultdict(int)-like object for unhashable components
 
     def __init__(self, otherdict):
-        self._data = []
-        for comp, count in otherdict.items():
-            self._data.append((comp, count))
+        # [(component, count)]
+        self._data = [item for item in otherdict.items()]
 
     def __getitem__(self, key):
-        for comp, count in self._data:
-            if comp == key:
+        for component, count in self._data:
+            if component == key:
                 return count
         return 0
 
@@ -75,24 +74,24 @@ class _UtilityRegistrations(object):
     _regs_for_components = WeakKeyDictionary()
 
     @classmethod
-    def for_components(cls, comps):
+    def for_components(cls, components):
         # We manage these utility/subscription registrations as associated
         # objects with a weakref to avoid making any changes to
         # the pickle format
         try:
-            regs = cls._regs_for_components[comps]
+            regs = cls._regs_for_components[components]
         except KeyError:
             regs = None
         else:
             # In case the components have been re-initted, clear the cache
             # (zope.component.testing does this between tests)
-            if (regs._utilities is not comps.utilities
-                or regs._utility_registrations is not comps._utility_registrations):
+            if (regs._utilities is not components.utilities
+                or regs._utility_registrations is not components._utility_registrations):
                 regs = None
 
         if regs is None:
-            regs = cls(comps.utilities, comps._utility_registrations)
-            cls._regs_for_components[comps] = regs
+            regs = cls(components.utilities, components._utility_registrations)
+            cls._regs_for_components[components] = regs
 
         return regs
 
@@ -117,22 +116,24 @@ class _UtilityRegistrations(object):
         try:
             self._cache[provided][component] += 1
         except TypeError:
-            # Not hashable, and we have a dict. Switch to a list.
-            self._cache[provided] = _CacheList(self._cache[provided])
-            self._cache[provided][component] += 1
+            # The component is not hashable, and we have a dict. Switch to a strategy
+            # that doesn't use hashing.
+            prov = self._cache[provided] = _UnhashableComponentCounter(self._cache[provided])
+            prov[component] += 1
 
     def __uncache_utility(self, provided, component):
+        provided = self._cache[provided]
         # It seems like this line could raise a TypeError if component isn't
         # hashable and we haven't yet switched to _CacheList. However,
         # we can't actually get in that situation. In order to get here, we would
         # have had to cache the utility already which would have switched
         # the datastructure if needed.
-        count = self._cache[provided][component]
+        count = provided[component]
         count -= 1
         if count == 0:
-            del self._cache[provided][component]
+            del provided[component]
         else:
-            self._cache[provided][component] = count
+            provided[component] = count
         return count > 0
 
     def _is_utility_subscribed(self, provided, component):
--- a/src/zope/interface/registry.py
+++ b/src/zope/interface/registry.py
@@ -13,9 +13,12 @@
 ##############################################################################
 """Basic components support
 """
+from collections import defaultdict
+from weakref import WeakKeyDictionary
+
 try:
     from zope.event import notify
-except ImportError: #pragma NO COVER
+except ImportError: # pragma: no cover
     def notify(*arg, **kw): pass
 
 from zope.interface.interfaces import ISpecification
@@ -36,6 +39,128 @@ from zope.interface.declarations import providedBy
 from zope.interface.adapter import AdapterRegistry
 from zope.interface._compat import CLASS_TYPES
 from zope.interface._compat import STRING_TYPES
+
+
+class _CacheList(object):
+    # defaultdict(int)-like object for unhashable components
+
+    def __init__(self, otherdict):
+        self._data = []
+        for comp, count in otherdict.items():
+            self._data.append((comp, count))
+
+    def __getitem__(self, key):
+        for comp, count in self._data:
+            if comp == key:
+                return count
+        return 0
+
+    def __setitem__(self, component, count):
+        for i, data in enumerate(self._data):
+            if data[0] == component:
+                self._data[i] = component, count
+                return
+        self._data.append((component, count))
+
+    def __delitem__(self, component):
+        for i, data in enumerate(self._data):
+            if data[0] == component:
+                del self._data[i]
+                return
+        raise KeyError(component) # pragma: no cover
+
+
+class _UtilityRegistrations(object):
+
+    _regs_for_components = WeakKeyDictionary()
+
+    @classmethod
+    def for_components(cls, comps):
+        # We manage these utility/subscription registrations as associated
+        # objects with a weakref to avoid making any changes to
+        # the pickle format
+        try:
+            regs = cls._regs_for_components[comps]
+        except KeyError:
+            regs = None
+        else:
+            # In case the components have been re-initted, clear the cache
+            # (zope.component.testing does this between tests)
+            if (regs._utilities is not comps.utilities
+                or regs._utility_registrations is not comps._utility_registrations):
+                regs = None
+
+        if regs is None:
+            regs = cls(comps.utilities, comps._utility_registrations)
+            cls._regs_for_components[comps] = regs
+
+        return regs
+
+    @classmethod
+    def clear_cache(cls):
+        cls._regs_for_components.clear()
+
+    def __init__(self, utilities, utility_registrations):
+        # {provided -> {component: count}}
+        self._cache = defaultdict(lambda: defaultdict(int))
+        self._utilities = utilities
+        self._utility_registrations = utility_registrations
+
+        self.__populate_cache()
+
+    def __populate_cache(self):
+        for ((p, _), data) in iter(self._utility_registrations.items()):
+            component = data[0]
+            self.__cache_utility(p, component)
+
+    def __cache_utility(self, provided, component):
+        try:
+            self._cache[provided][component] += 1
+        except TypeError:
+            # Not hashable, and we have a dict. Switch to a list.
+            self._cache[provided] = _CacheList(self._cache[provided])
+            self._cache[provided][component] += 1
+
+    def __uncache_utility(self, provided, component):
+        # It seems like this line could raise a TypeError if component isn't
+        # hashable and we haven't yet switched to _CacheList. However,
+        # we can't actually get in that situation. In order to get here, we would
+        # have had to cache the utility already which would have switched
+        # the datastructure if needed.
+        count = self._cache[provided][component]
+        count -= 1
+        if count == 0:
+            del self._cache[provided][component]
+        else:
+            self._cache[provided][component] = count
+        return count > 0
+
+    def _is_utility_subscribed(self, provided, component):
+        try:
+            return self._cache[provided][component] > 0
+        except TypeError:
+            # Not hashable and we're still using a dict
+            return False
+
+    def registerUtility(self, provided, name, component, info, factory):
+        subscribed = self._is_utility_subscribed(provided, component)
+
+        self._utility_registrations[(provided, name)] = component, info, factory
+        self._utilities.register((), provided, name, component)
+
+        if not subscribed:
+            self._utilities.subscribe((), provided, component)
+
+        self.__cache_utility(provided, component)
+
+    def unregisterUtility(self, provided, name, component):
+        del self._utility_registrations[(provided, name)]
+        self._utilities.unregister((), provided, name)
+
+        subscribed = self.__uncache_utility(provided, component)
+
+        if not subscribed:
+            self._utilities.unsubscribe((), provided, component)
 
 
 @implementer(IComponents)
@@ -98,17 +223,7 @@ class Components(object):
                 return
             self.unregisterUtility(reg[0], provided, name)
 
-        subscribed = False
-        for ((p, _), data) in iter(self._utility_registrations.items()):
-            if p == provided and data[0] == component:
-                subscribed = True
-                break
-
-        self._utility_registrations[(provided, name)] = component, info, factory
-        self.utilities.register((), provided, name, component)
-
-        if not subscribed:
-            self.utilities.subscribe((), provided, component)
+        _UtilityRegistrations.for_components(self).registerUtility(provided, name, component, info, factory)
 
         if event:
             notify(Registered(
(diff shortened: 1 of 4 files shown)


jamadden:
There was some trailing whitespace in `test_registry.py` that got cleaned up. Adding [`?w=0` to the files link](https://github.com/zopefoundation/zope.interface/pull/48/files?w=0) may make the diff a bit easier to read.

jamadden:
Are there any interested reviewers (maybe @tseaver)?

tseaver:
I was just looking at it, but my knowledge of the internals has evaporated.  Maybe @jimfulton has a more recent swapfile?

jamadden:
I realize it looks like a lot of textual change, but hopefully what it's doing is actually pretty straightforward :) (The sketch in #46 is a somewhat simpler version.) 

In essence we're changing the O(n) loop in :
```python
   def registerUtility(...)
       subscribed = False
        for ((p, _), data) in iter(self._utility_registrations.items()): # BAD
            if p == provided and data[0] == component:
                subscribed = True
                break

        self._utility_registrations[(provided, name)] = component, info, factory
        self.utilities.register((), provided, name, component)

        if not subscribed:
            self.utilities.subscribe((), provided, component)
```

into the O(1) lookup:
```python
   def _is_utility_subscribed(self, provided, component):
        return self._cache[provided][component] > 0

   def registerUtility(...)
        subscribed = self._is_utility_subscribed(provided, component) # GOOD

        self._utility_registrations[(provided, name)] = component, info, factory
        self.utilities.register((), provided, name, component)

        if not subscribed:
            self.utilities.subscribe((), provided, component)
```

The greater textual changes come from two things. First, refactoring to be able to easily keep the cache maintenance in the same place as the actual registrations/unregistrations (hence `_UtilityRegistrations`) and second, dealing with components that can't be hashed (hence `_CacheList`...which could perhaps use a better name).

prsephton:
My 2c fwiw:

Strictly speaking its O(log(n)) for a dict.

There may be some further optimisation possible for example:
`
    def __uncache_utility(self, provided, component):
        # It seems like this line could raise a TypeError if component isn't
        # hashable and we haven't yet switched to _CacheList. However,
        # we can't actually get in that situation. In order to get here, we would
        # have had to cache the utility already which would have switched
        # the datastructure if needed.
        prov = self._cache[provided]
        count = prov[component]
        count -= 1
        if count == 0:
            del prov[component]
        else:
            prov[component] = count
        return count > 0
`
or
`
    def __cache_utility(self, provided, component):
        prov = self._cache[provided]       # may need try+except?
        try:
            prov[component] += 1
        except TypeError:
            # Not hashable, and we have a dict. Switch to a list.
            prov = self._cache[provided] = _CacheList(self._cache[provided])
            prov[component] += 1
`

Since components may be expected to be instances of zope.component.Interface, one might
define a default __hash__ function in the superclass to make all components hashable?  Just a thought.

prsephton:
Just realised the bit about "instance of Interface" is not necessarily true

jamadden:
> Strictly speaking its O(log(n)) for a dict.

It's implementation dependent. Hashtables, which is what CPython uses to implement dicts, are generally [O(1) for lookups](https://wiki.python.org/moin/TimeComplexity) if things are well distributed; bad hashing, of course, throws that all out the window.

jamadden:
>     def __uncache_utility(self, provided, component):

Yup, we can save one dict lookup there in every case. Good catch, thanks.

>     def __cache_utility(self, provided, component):

Here, switching between the dict strategy and the list strategy shouldn't be common. And when it does occur, an extra dict lookup will be dominated by the cost to construct the list. So saving a dict lookup is probably not a win. It may make the code slightly clearer, though.

prsephton:
I stand corrected- I thought it was a red/black tree.

jamadden:
Feedback incorporated.

It looks like one of the AppVeyor builds is hung, been setting at the install step for 45 minutes. Everything else is green (though I had to restart one of the Travis Mac builds, those also seem to hang sometimes).

jamadden:
That hung appveyor build was eventually terminated after an hour, but all the other appveyor builds were green.

jamadden:
Maybe one of the people who +1'd #46 would like to review or suggest someone who might? Perhaps @icemac?

icemac:
@jamadden I put it on my list.

jamadden:
@icemac Thanks!

src/zope/interface/registry.py
```diff
@@ -38,6 +41,129 @@ def notify(*arg, **kw): pass
...
+        except TypeError:
+            # The component is not hashable, and we have a dict. Switch to a strategy
+            # that doesn't use hashing.
+            prov = self._cache[provided] = _UnhashableComponentCounter(self._cache[provided])
+            prov[component] += 1
+
+    def __uncache_utility(self, provided, component):
+        provided = self._cache[provided]
+        # It seems like this line could raise a TypeError if component isn't
+        # hashable and we haven't yet switched to _CacheList. However,
```
icemac: I think this `_CacheList` should be `_UnhashableComponentCounter`.

src/zope/interface/registry.py
```diff
@@ -38,6 +41,129 @@ def notify(*arg, **kw): pass
 from zope.interface._compat import STRING_TYPES
 
 
+class _UnhashableComponentCounter(object):
```
icemac: It would be nice to have a docstring telling when this class is used.

icemac:
I tested the branch in a mid-sized Zope-3 project having about 25 global utilities and 10 local utilities on a local site manager. There where no issues when running the tests and starting the server with a Data.fs created before switching to the branch.

I did not see any performance gains when running the tests but I think I have too little utilities and the (un-) registrations are only a very small part of the test run.

Besides the two little comments above: LGTM.

src/zope/interface/registry.py
jamadden: The next line has a comment for that purpose, in lieu of a docstring (since it's not a public class).

jamadden:
Thanks! I'll merge once the build goes green.

jamadden:
A PyPI release would be appreciated when someone has the time.

icemac:
@jamadden I'll have a look. But it is a bit more complicated than usual because direct commits to the master branch are forbidden in this repository.

icemac:
@jamadden Released as 4.3.0 to [PyPI](https://pypi.org/project/zope.interface/4.3.0/). Merge of the release branch (#52) currently pending.

mgedmin:
Welp, does GitHub branch protection prevent people who aren't repository admins from pushing version bump commits directly to master?  That's nasty!

Here's what the Settings page gives me:

ekrano nuotrauka is 2016-08-31 14-56-50

I can uncheck the "Require status checks to pass before merging" checkbox.  Is there maybe some way to have this requirement for GitHub's green Merge button without disabling direct git push'es?

mgedmin:
> I can uncheck the "Require status checks to pass before merging" checkbox.

I have done that thing because I want to encourage frequent releases.

jamadden:
Unfortunately this pull had a regression so it looks like we'll need one of those frequent releases 😄  See #53 and the corresponding PR in #54

icemac:
@jamadden When the CI tests are green I can cut a release.

jamadden:
Thanks @icemac. They turned green remarkably fast this time (maybe it's a good time for the mac farm).

icemac:
@mgedmin Your change in the repository configuration allowed me to push directly to `master`. I am fine with that but not sure if this was what you intended.

mgedmin:
> I am fine with that but not sure if this was what you intended.

Yes, that's what I intended.  Pull requests for commits like pre-/post-release version bump make no sense to me.

strichter:
This PR is absolute insane! It leaks objects and memory like crazy!

There is no testing support for clearing _UtilityRegistrations._cache anywhere.

In addition, when registering a ZODB-based utility (pretty common use case), then the cache leaves stale persistent objects around.

jamadden:
I appreciate your concern. Can you elaborate on the issues you've been seeing?

As a data point, we've been using this in several large production ZODB/zope3 sites since its release (well, since its followup regression-fix release 4.3.2 on September 5th) with no memory or stale object issues that we've noticed. Our services get restarted maybe three times a month, but large numbers of utilities are registered and unregistered on a near daily basis (hence the need for this optimization).

jimfulton:
Sorry for coming late to this, but this changed caused pain for Shoobx tests and they've asked me to take a look.

src/zope/interface/registry.py
```diff
@@ -38,6 +41,129 @@ def notify(*arg, **kw): pass
...
+        for i, data in enumerate(self._data):
+            if data[0] == component:
+                del self._data[i]
+                return
+        raise KeyError(component) # pragma: no cover
+
+
+class _UtilityRegistrations(object):
+
+    _regs_for_components = WeakKeyDictionary()
```
jimfulton: Why is this connected to utility registries via a WeakKeyDictionary rather than just adding a ``_cache`` attribute to Components?  This indirection seems pretty weird because it distributes the state.

src/zope/interface/registry.py
```diff
@@ -38,6 +41,129 @@ def notify(*arg, **kw): pass
...
+        try:
+            regs = cls._regs_for_components[components]
+        except KeyError:
+            regs = None
+        else:
+            # In case the components have been re-initted, clear the cache
+            # (zope.component.testing does this between tests)
+            if (regs._utilities is not components.utilities
+                or regs._utility_registrations is not components._utility_registrations):
+                regs = None
```
jimfulton: This lazy operation is probably what's causing the test problems for shoobx.

If this registry was just an ordinary attribute of ``Components``, it would be reset in ``__init__``.

src/zope/interface/registry.py
jamadden: We didn't desire to change the pickle or object state in any way.

src/zope/interface/registry.py
jamadden: > This lazy operation is probably what's causing the test problems for shoobx.

Because objects are persisting in memory longer than they previously did? They're not accessed (the first time they *would* be, it notices `__init__` has been called and throws them away) so I wouldn't anticipate any problems with closed Connections or the like. Are the tests using `gc.get_objects()` or otherwise doing "leak" tracking?

jimfulton: Yes, they are doing leak tracking. I'm not familiar with the details.

jamadden: Adding `_UtilityRegistrations.for_components(self)` to the bottom of `Components.__init__` would probably take care of it then, at the potential cost of an extra object for Components instances that never actually have utility registrations.

jimfulton: Yup, although I'd prefer something that was more explicit about test cleanup. I'll experiment with this and their tests.

src/zope/interface/registry.py
jimfulton: FWIW (not much, since I no longer have a stake in this) I think it would have been better to change the state.  I think it would have been a lot cleaner and the evolution could have been done lazily via an accessor/property that populated data if it was missing.

src/zope/interface/registry.py
jimfulton: Nope.  Because it doesn't deal with persistent registries.

I have a hack that seems to fix the shoobx tests, but I don't like it.  Basically, if a Components sees it is cached in its __init__, it assumes it's being called from test setUp or tearDown.  This is nasty.

Amongst the things that are bad here:

- __init__ has 2 responsibilities, initialization and test cleanup.  There should be an explicit cleanup mechanism that deals both with the base registry and the cache (if we keep the cache the way it is).

- The complicated weakref lazy cache this PR introduced.  This should be changed to an attribute, with a mechanism to create it for old instances or a _v_ attr.  The latter might be necessary if persistent registrations might somehow be made in the non-persistent registry, but I'm guessing this isn't a problem.

PR #61, opened 2016-11-15
Fix C optimizations broken on Py3k.

### LINKED ISSUE #60: C optimizations
When we are loading zope.interface on our projects the coptimizations are not loaded saying:

ImportError: No module named '_zope_interface_coptimizations'

If we change the import to zope.interface._zope_interface_coptimizations it works.

Python 3.5

tseaver:
- The bare import of `_zope_c_optimizations` prevented them from being   used.

- After enabling them via absolute imports, they would fail due to http://bugs.python.org/issue15657.

Fixes #60.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changes
 4.3.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix C optimizations broken on Py3k.  See the Python bug at:
+  http://bugs.python.org/issue15657
+  (https://github.com/zopefoundation/zope.interface/issues/60)
 
 
 4.3.2 (2016-09-05)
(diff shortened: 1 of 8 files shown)


jamadden:
The change itself looks good to me, to the extent that I understand the referenced Python issue, and AppVeyor passed. 

Travis has been hung up all this time on the Mac builds (which haven't been passing for quite some time anyway). I've cleared the backlog for this repository, and we can see if the linux builds pass (they did), but I'm not hopeful Travis will go green anytime soon.

jamadden:
The other branch referenced in #60 added a partial test case to make sure we get optimizations when we're expecting them. Is that worth doing here?

tseaver:
@jamadden I've added tests asserting that the optimizations are / are not bound to the expected names.

jamadden:
@tseaver Thank you! LGTM. Again, Appveyor passed, and I've done the Travis dance to clear out the last build (hung on the Macs), but the build for this one hasn't started yet.

jamadden:
The portion of travis that's going to go green has gone green.

PR #65, opened 2016-12-02
Removed docstrings to avoid publishing in Zope2 environments.

mauritsvanrees:
This is part of [PloneHotfix20161129](https://plone.org/security/hotfix/20161129/unauthorized-disclosure-of-site-configuration).

This is a very minor part of it: strangely this was only needed on Plone 4.1. For the current releases 4.3 and 5.0 a different patch, targeting z3c.form widgets, was used, which somehow was not enough on 4.1.

A possible line of 'attack' (information disclosure) would be:

- You have a z3c.form on `localhost/plone/@@discussion-settings`
- That form has a widget at `/++widget++captcha`
- You can get at the Interface of this widget using `/terms/field/interface`
- And then you can access various methods, like `/setTaggedValue?tag=cake&value=lovely` and `/namesAndDescriptions`.

Core Plone has a [test](https://github.com/plone/Products.CMFPlone/blob/5.0.6/Products/CMFPlone/tests/testSecurity.py#L71-L76) for this situation. The z3c.form patch that I mentioned actually results in a 302 redirect to the login form in the above scenario. Without that, you would get a 200. With the changes of this pull request you would get a 404 NotFound.

Like I said, this pull request is not strictly needed on current Plone releases. But there may be other ways, also outside of Plone in core Zope2 or CMF, to get at the Interface through a url, which would give an attacker lots of information about the Interface.

If this pull request is accepted, I should probably port this to branches 3.6 and 3.7.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 4.3.3 (unreleased)
 ------------------
 
+- Removed docstrings to avoid publishing in Zope2 environments.
+
 - Fix C optimizations broken on Py3k.  See the Python bug at:
   http://bugs.python.org/issue15657
   (https://github.com/zopefoundation/zope.interface/issues/60)
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -66,27 +66,27 @@ class Element(object):
         self.__tagged_values = {}
 
     def getName(self):
-        """ Returns the name of the object. """
+        # Returns the name of the object.
         return self.__name__
 
     def getDoc(self):
-        """ Returns the documentation for the object. """
+        # Returns the documentation for the object.
         return self.__doc__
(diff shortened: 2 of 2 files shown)


tseaver:
It feels wrong to "damage" this library (e.g., making interface objects less useful at the REPL) to address a Zope2-specific security issue:  ISTM it would be better to have Zope2 take on the burden of removing the docstrings at server startup time (e.g., by migrating the hotfix code into Zope2 itself).

tseaver:
Also, I'm removing the "bug" label:  the presence of docstrings cannot be considered a bug, outside of the concerns of one downstream consumer of this project.

mauritsvanrees:
Okay, I am opening pull requests in Zope2 for this. Thanks for the feedback.

PR #73, opened 2017-01-05
Expose IComponent documentation.

### LINKED ISSUE #70: document the zope.interface.registry.Components component registry
The `Components` class has no clear documentation within `zope.interface`. The closest I can find is the `AdapterRegistry` but that misses several aspects of the `Components` class. This is, of course, because the `Components` class is higher level than the `AdapterRegistry` despite having a very similar API.

Ideally this would include examples or explanation of why you would use `Components` instead of `AdapterRegistry` as well as exposing the `IComponents`, `IComponentLookup` and `IComponentRegistry` interfaces.

Related: #41

jamadden:
Fixes #70


jamadden:
I re-enabled the webhook for RTD and built the latest version of the docs. The HTML builds successfully, but latex and maybe others do not. Can someone with control over https://readthedocs.org/projects/zopeinterface/ (looks like @tseaver) take a look and maybe just disable the latex/epub/pdf generation?

jamadden:
Thanks for the reviews!

PR #75, opened 2017-02-16
Don't cache lookup failures

tschorr:
`z.i.adapter.LookupBase._getcache` will only use the `name` argument as a cache key if `name` evaluates to True (https://github.com/zopefoundation/zope.interface/blob/3.6/src/zope/interface/adapter.py#L314). If a client uses a name that evaluates to False (in addition to the correct interface) to lookup an unnamed utility, the utility is not found, but the default marker ends up in the cache and calls to `lookup` will subsequently return it even as a result of correct queries (i.e. queries with no name provided, `name` then evaluates to the default u'') even though the utility exists and could be found using `_uncached_lookup()`.

Example:
https://github.com/eea/eea.facetednavigation/commit/92390c6a9b0f1d3bf748a565d9c4a1f743c02694#diff-baadd81b5599e2f44501742288141f51L247

Here, `queryUtility(ISolrSearch, name=None)` is used to lookup collective.solr's (unnamed) Search utility. The lookup fails, but the result of the query nevertheless gets cached. As a consequence subsequent (correct) queries using `queryUtility(ISolrSearch)` (without the name) will return the default marker.

Suggestion: Do not cache lookup failures. An alternative would be to fix the condition to use a broader range of names as cache key in LookupBase._getcache().

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@
 3.6.8 (unreleased)
 ------------------
 
+- Don't cache lookup failures.
 
 3.6.7 (2011-08-20)
 ------------------
@@ -16,7 +17,7 @@
 
 - LP #570942:  Now correctly compare interfaces  from different modules but
   with the same names.
-  
+
   N.B.: This is a less intrusive / destabilizing fix than the one applied in
   3.6.3:  we only fix the underlying cmp-alike function, rather than adding
   the other "rich comparison" functions.
@@ -202,7 +203,7 @@ New Features
 ++++++++++++
 
 - The adapter-lookup algorithim was refactored to make it
-  much simpler and faster.  
+  much simpler and faster.
 
   Also, more of the adapter-lookup logic is implemented in C, making
   debugging of application code easier, since there is less
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -38,7 +38,7 @@ static PyTypeObject *Implements;
 
 static int imported_declarations = 0;
 
-static int 
+static int
 import_declarations(void)
 {
   PyObject *declarations, *i;
@@ -46,7 +46,7 @@ import_declarations(void)
   declarations = PyImport_ImportModule("zope.interface.declarations");
   if (declarations == NULL)
     return -1;
-  
+
   BuiltinImplementationSpecifications = PyObject_GetAttrString(
                     declarations, "BuiltinImplementationSpecifications");
   if (BuiltinImplementationSpecifications == NULL)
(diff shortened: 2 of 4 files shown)


icemac:
@tschorr Do you solve here the issue I reported as zopefoundation/zope.component#24?

tschorr:
@icemac looking at the call stack I'd say yes there is a good chance that https://github.com/zopefoundation/zope.component/issues/24 refers to the same problem.

pbauer:
Use the magic ?w=1 (https://github.com/zopefoundation/zope.interface/pull/75/files?w=1) to see the changeset without distracting whitespace-changes (there were plenty of obsolete whitespaces that got auto-removed by vim).

mauritsvanrees:
LGTM, and should solve a problem that we are sometimes seeing.
Should probably be done on the 3.7 and master branches as well. 

Tests pass locally, both with and without the C extensions.
But no travis tests are run, because that has not been setup on the 3.6 branch. I am about to create a pull request for that.

tseaver:
I have to ask: why are you working on making changes to a five-and-a-half-year-old branch?

Rotonen:
@mauritsvanrees looks legit on the C front on a quick visual scan, you +1 on the returned result and leave the at that branch now useless `Py_None` to be gobbled up by the GC at its convenience.

```
$HOME/.homebrew/opt/llvm/bin/scan-build \
    /usr/bin/gcc \
    -I$HOME/.homebrew/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/include/python2.7/ \
    -c _zope_interface_coptimizations.c
scan-build: Using '<redacted>/.homebrew/Cellar/llvm/4.0.0_1/bin/clang-4.0' for static analysis
scan-build: Removing directory '/var/folders/x4/7q6487553jd29w71f88t9ktw0000gn/T/scan-build-2017-04-22-183232-37060-1' because it contains no reports.
scan-build: No bugs found.
```

Nothing flares up on LLVM scan-build in general either.

tschorr:
@tseaver because zope.interface 3.6.7 is the version used by Plone 5.0.6, where this problem came up.

mauritsvanrees:
I went ahead and rebased this on the 3.6 branch, which now has Travis/appveyor support, so we can see what the automated checkers have to say.

tseaver:
@mauritsvanrees Because passing anything but ext for `name` is a programming error, I'd like to propose a different fix:  instead of monkeying with the caching, let's just to a typecheck on `name` and raise a ValueError.

mauritsvanrees:
@tseaver That seems to make sense.

tseaver:
Closing in favor of the solution in #82.

PR #77, opened 2017-03-20
Fix sentence in README

niels-bom-md:
Classes (factories) _implement_, objects _provide_. After first carefully explaining the difference, the docs mix them up.


icemac:
@niels-bom-md Thank you for fixing this.

PR #78, opened 2017-04-21
CI setup for manylinux wheels on tagged commit

fgregg:
This is basically the same as how we do it for [BTrees](https://github.com/zopefoundation/BTrees/commit/d1e33b1ebdffbe84b7e82eef920e15821f43019d).


fgregg:
@tseaver what do you think about cutting a new release?

tseaver:
I'll roll a 4.4.0 today (since we added Python 3.6 support).

PR #79, opened 2017-04-22
Added Travis, appveyor, and tox for 3.6 branch.

mauritsvanrees:
Copied from master branch, but removed the Python 3 stuff, which is unsupported on the 3.6 branch. Well, 3.1 seems supported, but that is very old, and on 3.3 I got errors locally.

Also removed 'docs' from the tox envs, as that proved tricky to get to work, and has changed much.

Removed `TODO_test_classProvides_fails_for_odd_class` test method, which would always fail but be ignored due to its name, except that the coverage tests would *not* ignore it, leading to a failure.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[report]
+exclude_lines =
+    pragma: no cover
+    class I[A-Z]\w+\((Interface|I[A-Z].*)\):
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,16 @@
+*.egg-info
+*.pyc
+*.so
+__pycache__
+.coverage
+.installed.cfg
+nosetests.xml
+coverage.xml
+.eggs/
+.tox/
+bin/
+build/
+eggs/
+develop-eggs/
+docs/_build/
+parts/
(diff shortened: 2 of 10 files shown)


mauritsvanrees:
On PyPy Travis shows an error:

```
 File "/home/travis/build/zopefoundation/zope.interface/src/zope/interface/__init__.py", line 55, in <module>
    from zope.interface.interface import Interface, _wire
  File "/home/travis/build/zopefoundation/zope.interface/src/zope/interface/interface.py", line 707, in <module>
    Interface = InterfaceClass("Interface", __module__ = 'zope.interface')
RuntimeError: maximum recursion depth exceeded
```

I think we can just ignore PyPy on this branch. I will remove it from the travis yml file.

tseaver:
I have to ask:  why are you working on making changes to a five-and-a-half-year-old branch?

mauritsvanrees:
Because that is what current Plone is still using, both 4.3 and 5.0. They are on Zope2 branch 2.13 which [still uses that](http://download.zope.org/Zope2/index/2.13.26/versions.cfg).
Plone 5.1 (in beta) is pinning zope.interface 4.2.0 (though still Zope2 2.13).

mauritsvanrees:
Thank you!

PR #82, opened 2017-05-04
Disallow non-text name passed to adapter registry methods.

tseaver:
Prevents corruption of lookup cache.

Alternative to fix proposed in #75 (3.6 branch).


tseaver:
@mauritsvanrees This is the variant of #81 for the 3.6 branch.

PR #54, opened 2016-08-31
Support Components subclasses that are not hashable.

jamadden:
Fixes #53.

This depends on the `id` of the instance staying stable and not being reused while the instance is extant. It uses a manual weakref and cleanup function.

Issue #53, opened 2016-08-31
regression in 4.3.0


mmerickel:
4.3.0 broke almost every single test in pyramid's test suite and appears to be related to https://github.com/zopefoundation/zope.interface/issues/46

See https://travis-ci.org/Pylons/pyramid/jobs/156458742 for an example or a snippet pasted below (all the failures look the same):

```
ERROR: test_authenticated_userid_fails_cleaning_as_Authenticated (pyramid.tests.test_authentication.TestCallbackAuthenticationPolicyDebugging)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Pylons/pyramid/pyramid/tests/test_authentication.py", line 12, in setUp
    self.config = testing.setUp()
  File "/home/travis/build/Pylons/pyramid/pyramid/testing.py", line 470, in setUp
    config._set_settings(settings)
  File "/home/travis/build/Pylons/pyramid/pyramid/config/settings.py", line 15, in _set_settings
    self.registry.settings = settings
  File "/home/travis/build/Pylons/pyramid/pyramid/registry.py", line 99, in _set_settings
    self.registerUtility(settings, ISettings)
  File "/home/travis/build/Pylons/pyramid/.tox/py27/lib/python2.7/site-packages/zope/interface/registry.py", line 227, in registerUtility
    _UtilityRegistrations.for_components(self).registerUtility(provided, name, component, info, factory)
  File "/home/travis/build/Pylons/pyramid/.tox/py27/lib/python2.7/site-packages/zope/interface/registry.py", line 82, in for_components
    regs = cls._regs_for_components[components]
  File "/opt/python/2.7.9/lib/python2.7/weakref.py", line 315, in __getitem__
    return self.data[ref(key)]
TypeError: unhashable type: 'Registry'
```

mmerickel:
This issue presents itself in production as well. Not just in the test suite.

jamadden:
Pyramid uses multiple-inheritance [to subclass `Components` *and* `dict`](https://github.com/Pylons/pyramid/blob/master/pyramid/registry.py#L19), hence leading to this issue. I guess we'll have to figure out another way to associate the necessary data.

icemac:
Fix released in 4.3.1.

icemac:
@mmerickel Please test using 4.3.1 in Pyramid.

jamadden:
[...]

PR #86, opened 2017-06-08
Partially revert #84 in order to fix #85

### LINKED ISSUE #85: 4.4.1 breaks pickling of PersistentComponents that have registrations
I [figured out what I was worried about](https://github.com/zopefoundation/zope.interface/pull/84#discussion_r116334080) in regards to #84.

The changes in 4.4.1 break the tests for zc.catalog (but it should be very simple to reproduce outside of those tests). All versions of Python now fail with a variation on this error:
```
  File "<doctest extentcatalog.rst[23]>", line 1, in <module>
    root = makeRoot()
  File "<doctest extentcatalog.rst[13]>", line 12, in makeRoot
    transaction.commit()
...
  File "//py27/lib/python2.7/site-
    return self._dump(meta, obj.__getstate__())
  File "//py27/lib/python2.7/site-packages/ZODB/serialize.py", line 428, in _dump
    self._p.dump(state)
  File "//.tox/py27/bin/../lib/python2.7/copy_reg.py", line 70, in _reduce_ex
TypeError: can't pickle function objects
```

Or `_pickle.PicklingError: Can't pickle <class 'function'>: attribute lookup builtins.function failed`.

Coercing Python 3 to use the Python implementation of ZODB pickle gets us a better error message:
```
  File "//py36/lib/python3.6/site-packages/zodbpickle/pickle_3.py", line 694, in save_global
    (obj, module, name))
_pickle.PicklingError: Can't pickle <function _UtilityRegistrations.__init__.<locals>.<lambda> at 0x11073c598>: it's not found as zope.interface.registry.<lambda>
```

Debugging the error shows that `state` is `{'data': {'components': <PersistentComponents>}}`, and further debugging shows that we're trying to pickle `PersistentComponents.__dict__`, leading to this error:

```
(Pdb) pp state
{'data': {'components': <PersistentComponents >}}
(Pdb) pp self._p.memo.copy()[id(state['data']['components'].__dict__)]
(7,
 {'__bases__': (<BaseGlobalComponents base>,),
  '__name__': '',
  '_adapter_registrations': {},
  '_handler_registrations': [],
  '_subscription_registrations': [],
  '_utility_registrations': {(<InterfaceClass zope.intid.interfaces.IIntIds>, ''): (<zope.intid.IntIds object at 0x10a9c42e8>,
                                                                                    '',
                                                                                    None)},
  '_v_utility_registrations_cache': <zope.interface.registry._UtilityRegistrations object at 0x10a9d6ac8>,
  'adapters': <zope.component.persistentregistry.PersistentAdapterRegistry object at 0x10a972eb8>,
  'utilities': <zope.component.persistentregistry.PersistentAdapterRegistry object at 0x10a9c40b8>})
```

The `makeRoot` function is very simple:
```pycon
>>> def makeRoot():
...     db = DB()
...     conn = db.open()
...     root = conn.root()
...     site_manager = root['components'] = (
...         zope.component.persistentregistry.PersistentComponents())
...     site_manager.__bases__ = (zope.component.getGlobalSiteManager(),)
...     site_manager.registerUtility(
...         zope.intid.IntIds(family=btrees_family),
...         provided=zope.intid.interfaces.IIntIds)
...     setSiteManager(site_manager)
...     transaction.commit()
...     return root
```

Reverting any tested python to zope.interface 4.4.0 fixes the problem.

The problem is that `PersistentComponents` *isn't* actually a `Persistent` object; only its `.adapters` and `.utilities` are, so the magic that's supposed to elide `_v` attributes never happens:

```python
class PersistentComponents(Components):

    def _init_registries(self):
        self.adapters = PersistentAdapterRegistry()
        self.utilities = PersistentAdapterRegistry()
```

Suggestions on how to tackle this?

jamadden:
But add testing to be sure it doesn't break again, and extra suspenders to ensure the issues that #84 was trying to deal with doesn't show up either.

FIxes #85

### DIFF
--- a/src/zope/interface/registry.py
+++ b/src/zope/interface/registry.py
@@ -14,6 +14,8 @@
 """Basic components support
 """
 from collections import defaultdict
+import functools
+import weakref
 
 try:
     from zope.event import notify
@@ -69,6 +71,53 @@ class _UnhashableComponentCounter(object):
 
 
 class _UtilityRegistrations(object):
+
+    # Strong reference {id(components): _UtilityRegistrations}
+    _regs_for_components = {}
+    # Weak reference {id(components): components}, used for cleanup.
+    _weakrefs_for_components = {}
+
+    @classmethod
+    def for_components(cls, components):
+        # We manage these utility/subscription registrations as associated
+        # objects with a weakref to avoid making any changes to
+        # the pickle format. They are keyed off the id of the component because
+        # Components subclasses are not guaranteed to be hashable.
+        key = id(components)
+        try:
+            regs = cls._regs_for_components[key]
+        except KeyError:
+            regs = None
+        else:
+            # In case the components have been re-initted, clear the cache
+            # (zope.component.testing does this between tests, which calls Components.__init__,
+            # so we should typically not get here)
+            if (regs._utilities is not components.utilities
+                or regs._utility_registrations is not components._utility_registrations):
+                regs = None # pragma: no cover
+
+        if regs is None:
+            regs = cls(components.utilities, components._utility_registrations)
+            cls._regs_for_components[key] = regs
+
+            if key not in cls._weakrefs_for_components:
+                cleanup = functools.partial(cls._cleanup_for_components, key)
+                cls._weakrefs_for_components[key] = weakref.ref(components, cleanup)
+
+        return regs
+
+    @classmethod
+    def _cleanup_for_components(cls, key, *args):
+        cls._weakrefs_for_components.pop(key, None)
+        cls._regs_for_components.pop(key, None)
+
+    @classmethod
+    def reset_for_components(cls, components):
+        cls._cleanup_for_components(id(components))
+
+    @classmethod
+    def clear_cache(cls):
+        cls._regs_for_components.clear()
 
     def __init__(self, utilities, utility_registrations):
         # {provided -> {component: count}}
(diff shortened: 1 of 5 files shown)


src/zope/interface/registry.py
```diff
@@ -70,6 +72,53 @@ def __delitem__(self, component):
...
+    # Weak reference {id(components): components}, used for cleanup.
+    _weakrefs_for_components = {}
+
+    @classmethod
+    def for_components(cls, components):
+        # We manage these utility/subscription registrations as associated
+        # objects with a weakref to avoid making any changes to
+        # the pickle format. They are keyed off the id of the component because
+        # Components subclasses are not guaranteed to be hashable.
+        key = id(components)
```
mgedmin: This scares me: you have to be really careful to avoid problems due to object ID reuse.

Why not use a weakref.WeakKeyDictionary?

jamadden: The initial version of this patch did use a WeakKeyDictionary, but [it turned out](https://github.com/zopefoundation/zope.interface/issues/53) that there are Components subclasses that aren't hashable (the aforementioned Pyramid registry that also subclasses dict). So we had to [revert that and use id instead](https://github.com/zopefoundation/zope.interface/pull/54).

mgedmin: TIL WeakKeyDictionary wants keys to be hashable.  :(

jimfulton:
I don't think I'll find time to review this. I defer.

I will say that this framework has become too complex, IMO, and this feels like more of the same.

jamadden:
Replaced by #92.

Issue #87, opened 2017-06-08
Dev: Coverage reporting and test counts in inconsistent shape; enable coveralls?


jamadden:
Running `tox -e coverage` reports 100% coverage:

```
Name                                  Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------
zope/interface/__init__.py               30      0      0      0   100%
zope/interface/_compat.py                 3      0      0      0   100%
zope/interface/adapter.py               452      0    219      0   100%
zope/interface/advice.py                 66      0     28      0   100%
zope/interface/common/__init__.py         0      0      0      0   100%
zope/interface/common/idatetime.py        9      0      0      0   100%
zope/interface/common/interfaces.py      43      0      0      0   100%
zope/interface/common/mapping.py          3      0      0      0   100%
zope/interface/common/sequence.py         3      0      0      0   100%
zope/interface/declarations.py          340      0     92      0   100%
zope/interface/document.py               59      0     30      0   100%
zope/interface/exceptions.py             21      0      0      0   100%
zope/interface/interface.py             356      0    122      0   100%
zope/interface/interfaces.py             28      0      8      0   100%
zope/interface/registry.py              381      0    143      0   100%
zope/interface/ro.py                     20      0      8      0   100%
zope/interface/verify.py                 50      0     26      0   100%
---------------------------------------------------------------------------------
TOTAL                                  1864      0    676      0   100%
----------------------------------------------------------------------
Ran 874 tests in 1.308s
```

However, that's not at all what `coverage run -m zope.testrunner --test-path=src && coverage report` shows, notably all of the `_` prefixed modules and test modules are left out of the previous report:

```
Name                                                        Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------------------------------------------
src/zope/interface/__init__.py                                 30      0      0      0   100%
src/zope/interface/_compat.py                                  36     16     12      3    52%   36-49, 54-56, 60, 62, 19->36, 52->54, 59->60
src/zope/interface/_flatten.py                                 14     14      4      0     0%   14-35
src/zope/interface/adapter.py                                 461      8    221      1    99%   255, 295-297, 413-414, 449-450, 253->255
src/zope/interface/advice.py                                   76      3     34      2    95%   31-32, 100, 99->100, 191->193
src/zope/interface/common/__init__.py                           0      0      0      0   100%
src/zope/interface/common/idatetime.py                          9      0      0      0   100%
[...]

PR #89, opened 2017-06-09
Enable coveralls

jamadden:
Test coverage is not at 100% yet so it should fail this PR (I'll address coverage in a future PR; doing so now would have a small conflict with #86).

Also enable pip caching since we are installing more deps.

Update PyPy versions to latest available for speed and because the old version of PyPy3 can't install newer pip and coverage, etc.

Partially addresses #87

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
         - os: linux
           python: 3.6
         - os: linux
-          python: pypy-5.6
+          python: pypy-5.6.0
         - os: linux
           python: pypy3.3-5.5-alpha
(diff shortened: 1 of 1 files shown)


icemac:
@jamadden According to [travis-ci#6727](https://github.com/travis-ci/travis-ci/issues/6727#issuecomment-307186234) there should even be a `pypy3.3-5.5-alpha`.

jamadden:
Cool, I hadn't seen that list yet, I'll update. In my experience PyPy 5.6 is a big step up from 5.4 (PyPy3, not so much, until you get to the 3.5 beta release).

PR #90, opened 2017-06-09
100% coverage

jamadden:
Well, as close as possible without conflicting with #86.

This is almost entirely trivial changes:

- Make dummy methods that should never be called either raise NotImplementedError or call self.fail() in tests.
- Standardize the no-coverage pragma
- Add a few pragmas where coverage varies across versions/implementations, mostly for the sake of tox (travis should hit them all).

Also convert the tox command to use the `coverage` command and drop use of `nosetests`. This also drops generating the .xml coverage reports by default.

### DIFF
--- a/src/zope/interface/tests/test_advice.py
+++ b/src/zope/interface/tests/test_advice.py
@@ -221,7 +221,7 @@ class Test_determineMetaclass(unittest.TestCase):
         self.assertEqual(self._callFUT((Meta, type)), Metameta)
 
     @_skip_under_py2
-    def test_meta_of_class_py3k(self):
+    def test_meta_of_class_py3k(self): # pragma: no cover (tox runs coverage on Python 2)
         # Work around SyntaxError under Python2.
         EXEC = '\n'.join([
         'class Metameta(type):',
@@ -249,7 +249,7 @@ class Test_determineMetaclass(unittest.TestCase):
         self.assertEqual(self._callFUT((A, B,)), Meta_B)
 
     @_skip_under_py2
-    def test_multiple_in_hierarchy_py3k(self):
+    def test_multiple_in_hierarchy_py3k(self): # pragma: no cover (tox runs coverage on Python 2)
         # Work around SyntaxError under Python2.
         EXEC = '\n'.join([
         'class Meta_A(type):',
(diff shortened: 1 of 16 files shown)


jamadden:
I have no problem with that, seems like it should do the job. Should it be a follow-up PR? This one's already dropping nose and XML reporting.

src/zope/interface/tests/test_advice.py
```diff
@@ -268,7 +249,7 @@ class B(type):
         self.assertEqual(self._callFUT((A, B,)), Meta_B)
 
     @_skip_under_py2
-    def test_multiple_in_hierarchy_py3k(self):
+    def test_multiple_in_hierarchy_py3k(self): # pragma: no cover (tox runs coverage on Python 2)
```
mgedmin: I wonder if you can add `@_skip_under_py2` to .coveragerc and avoid this repetition.  (Or, even better, merge py2+py3 coverage results.)

src/zope/interface/tests/test_declarations.py
```diff
@@ -968,11 +961,14 @@ def test_w_non_descriptor_aware_metaclass(self):
         from zope.interface.interface import InterfaceClass
         IFoo = InterfaceClass("IFoo")
         class MetaClass(type):
-            def __getattribute__(self, name):
+            def __getattribute__(cls, name):
                 # Emulate metaclass whose base is not the type object.
                 if name == '__class__':
                     return self
```
mgedmin: I think you want to `return cls` here.

src/zope/interface/tests/test_interface.py
```diff
@@ -1523,11 +1520,10 @@ def _errorsEqual(self, has_invariant, error_len, error_msgs, iface):
         e = []
         try:
             iface.validateInvariants(has_invariant, e)
+            self.fail("validateInvariants should always rais")
```
mgedmin: Typo: rais -> raise

src/zope/interface/tests/test_declarations.py
jamadden: Heh. Nice catch! The test still works because `self` is defined in the outer scope.

src/zope/interface/tests/test_advice.py
jamadden: Whatdyaknow, you *can* add `@_skip_under_py2` to the .coveragerc and it marks the whole method ignored. I assumed it would only mark that one line. I guess coverage is smarter than I thought it was!

@icemac proposed combining coverage and has an example to do so. I'll try to follow up with a PR for that.

PR #92, opened 2017-06-14
Fix #85 by implementing __reduce__ to omit _v_ attributes.

### LINKED ISSUE #85: 4.4.1 breaks pickling of PersistentComponents that have registrations

jamadden:
This is an alternative to #86 and was suggested by @mgedmin in #85.

There are tests to be sure this works for subclasses like Pyramid's Registry that extend dict too.

This doesn't change the pickle format, we were storing the output of `object.__reduce__` previously.

I've verified this fixes the zc.catalog issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,15 +4,17 @@ Changes
 4.4.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
-
+- Fix a regression storing
+  ``zope.component.persistentregistry.PersistentRegistry`` instances.
+  See `issue 85 <https://github.com/zopefoundation/zope.interface/issues/85>`_.
 
 4.4.1 (2017-05-13)
 ------------------
 
 - Simplify the caching of utility-registration data. In addition to
   simplification, avoids spurious test failures when checking for
-  leaks in tests with persistent registries.
+  leaks in tests with persistent registries. See `pull 84
+  <https://github.com/zopefoundation/zope.interface/pull/84>`_.
 
 - Raise ``ValueError`` when non-text names are passed to adapter registry
   methods:  prevents corruption of lookup caches.
--- a/src/zope/interface/registry.py
+++ b/src/zope/interface/registry.py
@@ -154,7 +154,24 @@ class Components(object):
     def __repr__(self):
         return "<%s %s>" % (self.__class__.__name__, self.__name__)
 
+    def __reduce__(self):
+        # Mimic what a persistent.Persistent object does and elide
+        # _v_ attributes so that they don't get saved in ZODB.
+        # This allows us to store things that cannot be pickled in such
+        # attributes.
+        reduction = super(Components, self).__reduce__()
+        # (callable, args, state, listiter, dictiter)
+        # We assume the state is always a dict; the last three items
+        # are technically optional and can be missing or None.
+        filtered_state = {k: v for k, v in reduction[2].items()
+                          if not k.startswith('_v_')}
+        reduction = list(reduction)
+        reduction[2] = filtered_state
+        return tuple(reduction)
+
     def _init_registries(self):
+        # Subclasses have never been required to call this method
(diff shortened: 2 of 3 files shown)


jamadden:
> lgtm, but don't hold me accountable if something else breaks ;)

Fair enough! Thanks.

I've tested with z3c.baseregistry, Pyramid, zc.catalog and nti.site.

It turns out nti.site reveals a regression introduced in 4.4.1 that's not fixed by this, but that's a separate PR.

PR #94, opened 2017-06-14
Fix the cache getting out of sync with _utility_registrations.

### LINKED ISSUE #93: 4.4.1 Regression: `Components._utilities`/`_utility_registrations` can get out of sync with the cache
The `_UtilityRegistrations` object makes a reference to the `Component`'s `_utilities` and `_utility_registrations` when it is created. The `Component` object holds a strong reference to the `_UtilityRegistrations`. Presumably the `_UtilityRegistrations` holds these two references to avoid holding a reference to the `Component` instance and thus introducing a cycle.

The problem introduced in 4.4.1 is that if either of those two instance variables are assigned to, the cache object is not updated. Previously, accessing the cache would compare the identity of the objects it held with the identity of the objects the Component held and reset if they changed. Now that doesn't happen.

The use case is that subclasses may want to change the identity of `_utility_registrations` in the `registerUtility` method. Specifically, we have a subclass that will change the `_utility_registrations` object to a BTree instead of a dict when it grows to a certain size.

Now we *could* update our subclass to have a dependency on >= 4.4.1 and knowledge of the fact that the cache is now an instance attribute that also needs to be updated, but I'd prefer to see it handled here so that the cache is always consistent and guaranteed to be so.

jamadden:
Fixes #93

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@ Changes
   ``zope.component.persistentregistry.PersistentRegistry`` instances.
   See `issue 85 <https://github.com/zopefoundation/zope.interface/issues/85>`_.
 
+- Fix a regression that could lead to the utility registration cache
+  of ``Components`` getting out of sync. See `issue 93
+  <https://github.com/zopefoundation/zope.interface/issues/93>`_.
+
 4.4.1 (2017-05-13)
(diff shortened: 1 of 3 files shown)


jamadden:
Thank you!

PR #95, opened 2017-06-14
Enable combined coverage reports under tox and branch coverage

### LINKED ISSUE #91: Dev: Have tox produce combined coverage reports
See https://github.com/zopefoundation/zope.interface/pull/90#pullrequestreview-43303790

jamadden:
Fixes #91

Remove many (or most) now-redundant 'pragma: no cover'.

A few minor test changes to get full branch coverage.

### DIFF
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -436,8 +436,7 @@ class implementer_only:
 def _implements(name, interfaces, classImplements):
     # This entire approach is invalid under Py3K.  Don't even try to fix
     # the coverage for this block there. :(
-    if PYTHON3: # pragma: no cover
-        raise TypeError('Class advice impossible in Python3')
+    assert not PYTHON3, "Class advice impossible in Python 3"
     frame = sys._getframe(2)
     locals = frame.f_locals
 
@@ -481,7 +480,7 @@ def implements(*interfaces):
     """
     # This entire approach is invalid under Py3K.  Don't even try to fix
     # the coverage for this block there. :(
-    if PYTHON3: # pragma: no cover
+    if PYTHON3:
         raise TypeError(_ADVICE_ERROR % 'implementer')
     _implements("implements", interfaces, classImplements)
 
@@ -509,7 +508,7 @@ def implementsOnly(*interfaces):
     """
     # This entire approach is invalid under Py3K.  Don't even try to fix
     # the coverage for this block there. :(
-    if PYTHON3: # pragma: no cover
+    if PYTHON3:
         raise TypeError(_ADVICE_ERROR % 'implementer_only')
     _implements("implementsOnly", interfaces, classImplementsOnly)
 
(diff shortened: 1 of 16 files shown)


src/zope/interface/declarations.py
```diff
@@ -436,8 +436,7 @@ def __call__(self, ob):
 def _implements(name, interfaces, classImplements):
     # This entire approach is invalid under Py3K.  Don't even try to fix
     # the coverage for this block there. :(
-    if PYTHON3: # pragma: no cover
-        raise TypeError('Class advice impossible in Python3')
+    assert not PYTHON3, "Class advice impossible in Python 3"
```
tseaver: Please don't convert `TypeError` to `AssertionError` (we only raise them in test code).

src/zope/interface/tests/test_odd_declarations.py
```diff
@@ -258,7 +258,7 @@ class C(A, B):
         c.c
 
         import sys
-        if sys.version[0] == '2': # This test only makes sense under Python 2.x
+        if sys.version_info[0] < 3: # This test only makes sense under Python 2.x
```
tseaver: Rather than testing `sys.version_info`, this is probably better as a conditional import:

```python
        try:
            from types import ClassType
        except ImportError:
            pass
        else:
            assert not isinstance(C, (type, ClassType))

```

src/zope/interface/declarations.py
tseaver: The rationale is that `python -O` suppresses assertions:  we don't want to remove that check in that case.

jamadden: This is a private implementation method for this module. It's callers *already* perform this same check and raise a TypeError before calling it. So the `if/raise` statement was dead code.

src/zope/interface/tests/test_odd_declarations.py
jamadden: Yeah, it probably is :) It started out as a copy of the doctest in odd.py, I just fixed this when I was globally fixing the same thing in other files.

src/zope/interface/declarations.py
tseaver: Yup, if you're sure. :)

jamadden: I'm sure. Even if we take the `assert` out and manually try to use the implementation method, we get hit with another explicit TypeError:
```pycon
Python 3.6.1 (default, Jun  1 2017, 07:41:42)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from zope.interface.declarations import _implements, classImplements
>>> class Foo(object):
...   _implements('implements', Interface, classImplements)
...
  File "<stdin>", line 2, in Foo
  File "//.tox/py36/lib/python3.6/site-packages/zope/interface/declarations.py", line 446, in _implements
    raise TypeError(name+" can be used only from a class definition.")
TypeError: implements can be used only from a class definition.
```

jamadden:
Thank you!

I'm going to push 4.4.2 to PyPI now unless there are objections.

Issue #98, opened 2017-08-02
Interface definitions with Python 3.5+ type hints raise InvalidInterface


avanov:
Hi!

The following definition

```python

class ICriterion(Interface):
    name: str = Attribute('name')
```

raises  an exception:

```
zope.interface.exceptions.InvalidInterface: Concrete attribute, __annotations__
```

Seems like it could be fixed by playing around https://github.com/zopefoundation/zope.interface/blob/50a9193fea1e5873c59bf0add2e1ddca80c1bcf4/src/zope/interface/interface.py#L355 , but before that I'd like to clarify the position of `zope.interface` regarding `PEP484` and `PEP544` - are they mutually exclusive or could complement each other?

tseaver:
- I think dropping `__annotatins__` (like `__locals__` and `__qualname__`) would be fine.
- PEP484 is kind of orthagonal to interfaces:  adopting it in a codebase which uses interfaces should be fine (but we won't do them, at least the inline version, while we support Python < 3.5).
- I don't know how interfaces fit with PEP544, except that they fall into its ["non-goals"](https://www.python.org/dev/peps/pep-0544/#non-goals). Personally, after nearly twenty years of using interfaces, I find ABCs utterly pointless.

avanov:
@tseaver thanks! I'll prepare a PR for `__annotations__`.

avanov:
Closing this.

PR #99, opened 2017-08-11
Drop __annotations__ in interface declarations

avanov:
This PR solves #98

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changes
 4.4.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Avoid exceptions when the ``__annotations__`` attribute is added to interface definitions with Python 3.x type hints.
+  See `issue 98 <https://github.com/zopefoundation/zope.interface/issues/98>`_.
 
 
 4.4.2 (2017-06-14)
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -352,9 +352,10 @@ class InterfaceClass(Element, InterfaceBase, Specification):
         # Make sure that all recorded attributes (and methods) are of type
         # `Attribute` and `Method`
         for name, attr in list(attrs.items()):
-            if name in ('__locals__', '__qualname__'):
+            if name in ('__locals__', '__qualname__', '__annotations__'):
                 # __locals__: Python 3 sometimes adds this.
                 # __qualname__: PEP 3155 (Python 3.3+)
+                # __annotations__: PEP 3107 (Python 3.0+)
                 del attrs[name]
                 continue
             if isinstance(attr, Attribute):
--- a/src/zope/interface/tests/test_interface.py
+++ b/src/zope/interface/tests/test_interface.py
@@ -460,6 +460,15 @@ class InterfaceClassTests(unittest.TestCase):
         self.assertEqual(inst.__bases__, ())
         self.assertEqual(inst.names(), ATTRS.keys())
 
+    def test_ctor_attrs_w___annotations__(self):
+        ATTRS = {'__annotations__': {}}
+        klass = self._getTargetClass()
+        inst = klass('ITesting', attrs=ATTRS)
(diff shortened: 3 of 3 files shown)


jamadden:
We probably need a test for this, similar to the way there's one for `__locals__` at https://github.com/zopefoundation/zope.interface/blob/master/src/zope/interface/tests/test_interface.py#L454

avanov:
@jamadden added and rebased

avanov:
@jamadden amended, thanks!

CHANGES.rst
```diff
@@ -4,7 +4,8 @@ Changes
 4.4.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Add support for general-purpose Python 3.x annotations / type-hints in interface definitions.
```
jamadden: I wonder if that may be overstating things a bit. As I understand it, this simply drops top-level `__annotations__` completely from the class at runtime. It allows Attribute values to thus be annotated in the source, but that's about it. For something that evaluates the AST (mypy?) that's probably OK, but for something that uses that data at runtime, like the stdlib `typing` module, that's not exactly "supported". (Indeed, `tpying.get_type_hints` raises an exception when passed an interface: `TypeError: <InterfaceClass __main__.IFoo> is not a module, class, method, or function.`)

In addition, methods also still completely lose their annotations (because they're represented by a custom type). E.g., this works with the released versions of zope.interface today:
```python
class IFoo(Interface):
   def thing(s:str) -> str:
       pass
```

But `IFoo['thing']` has no `__annotations__`.

Was your use-case mypy or something else?

avanov: Yes, it is mypy. Ideally, I'd like to have annotations that hold type information as well. I stumbled upon the case when having a type hint for an attribute declaration resulted in runtime error:
```python

class ICriterion(Interface):
    name: str = Attribute('name')
```

Even though it's not correct from mypy's perspective either (Attribute() is not an instance of string). So far I'd like to just have this combination of hints and interfaces present at the same time.

jamadden: Does mypy think that `adapter = IAdapter(context)` produces an *instance* of `IAdapter`, and hence tries to propagate type information from `IAdapter`? If so, I can't decide if that's clever or silly of it, especially since there's no `__init__` method defined...

If `name: str = Attribute('name')` produces errors in mypy, have you considered using the external files support (.pyi)? Does that have the same problem? (I guess you could use a union type Union[str,Attribute] if so.)

Note that I'm not trying to say I've changed my mind and consider this a bad idea, I'm just trying to (a) understand your use case and try to help solve it better and (b) make sure we don't set expectations for `__annotations__` that aren't actually met---perhaps a more clear change note would just explicitly say something about attributes and mypy.

tseaver: I would also prefer a less ambitious changelog entry, similar to the one which added `__qualname__` to the list of skipped attrs in version 4.0.0.  For reference, that messsage was:

```rst
Avoid exceptions due to tne new ``__qualname__`` attribute added in Python 3.3 (see PEP 3155 for rationale). 
```

avanov: @jamadden my current use case is introducing interfaces into a project with a 6-year history, which includes convincing the team that it has value and doesn't have extra magic and corner/edge cases. Therefore, being able to specify type hints in interface attributes is a good thing to have, even though it might not be correctly handled by mypy at the moment (I haven't run it against interfaces yet, though, I assume it won't infer the type correctly because there are no stubs for `zope.interfaces`). At least, it is good to be able to specify them as in regular python objects.

icemac:
What has to be done to get this PR ready for merge?
Does the contributor to have to sign a contributor agreement?

tseaver:
IMO, a contributor agreement is not required (the actual change is trivial, and the added test likewise).  I requested a change to the changelog entry.

avanov:
@tseaver how about this updated changelog entry?

tseaver:
@avanov Thank you for your work!

PR #102, opened 2017-09-14
Allow Registering and unregistering instance methods as listeners

### LINKED ISSUE #12: `unregisterAdapter` uses '==' while the underlying adapter registry uses 'is'
In https://bugs.launchpad.net/zope.interface/+bug/551049, @mgedmin reported:

> Try this:

```
class MyTestClass(PlacelessSetUp):

    def _handler(self, event):
        print "Got event!"

    def test(self):
        sm = zope.component.getGlobalSiteManager()
        sm.registerHandler(self._handler, required=(zope.interface.Interface, ))
        assert sm.unregisterHandler(self._handler, required=(zope.interface.Interface, ))
        zope.event.notify(object())
```

> The assertion will demonstrate that the global site manager thinks it successfully unregistered the adapter, yet the subsequent notify() will invoke our handler.
> 
> Why? Because self._handler creates a new bound method instance every time you write that expression. sm.unregisterHandler compares it to the previously registered factory object using '==' which returns true. But then the site manager invokes zope.interface.adapter.BaseAdapterRegistry.unsubscribe, which uses 'is' for identity checking and therefore thinks it got a different handler.
> 
> zope.component should use the same kind of check as the underlying implementation, because it now lies to the programmer about unregistering an adapter that really remains registered.

@d03cc followed up:

> Trying to reproduce it, I noticed that the code that does not do the proper unregistration is in zope.interface and not in zope.component

dwt:
Fixes #12 - registering instance methods as listeners doesn’t allow to easily unregister them as the registry tries to find the handler with 'is' but it should use '==' to allow the python BoundMethod wrapper (which is a new instance every time instance.$methodname is accessed).

### DIFF
--- a/src/zope/interface/tests/test_adapter.py
+++ b/src/zope/interface/tests/test_adapter.py
@@ -236,7 +236,7 @@ class BaseAdapterRegistryTests(unittest.TestCase):
         self.assertEqual(len(registry._subscribers), 2)
     
     def _instance_method_notify_target(self):
-        pass
+        self.fail("Example method, not intended to be called.")
     
     def test_unsubscribe_instance_method(self):
         IB0, IB1, IB2, IB3, IB4, IF0, IF1, IR0, IR1 = _makeInterfaces()
--- a/src/zope/interface/tests/test_adapter.py
+++ b/src/zope/interface/tests/test_adapter.py
@@ -234,6 +234,17 @@ class BaseAdapterRegistryTests(unittest.TestCase):
         registry.subscribe([IB1], None, orig)
         registry.unsubscribe([IB1], None, nomatch) #doesn't raise
         self.assertEqual(len(registry._subscribers), 2)
+    
+    def _instance_method_notify_target(self):
+        pass
+    
+    def test_unsubscribe_instance_method(self):
+        IB0, IB1, IB2, IB3, IB4, IF0, IF1, IR0, IR1 = _makeInterfaces()
+        registry = self._makeOne()
+        self.assertEqual(len(registry._subscribers), 0)
+        registry.subscribe([IB1], None, self._instance_method_notify_target)
+        registry.unsubscribe([IB1], None, self._instance_method_notify_target)
+        self.assertEqual(len(registry._subscribers), 0)
 
 
 class LookupBaseFallbackTests(unittest.TestCase):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changes
 4.4.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Allow registering and unregistering instance methods as listeners.
+  See `issue 12 <https://github.com/zopefoundation/zope.interface/issues/12>`_
+  and `PR 102 <https://github.com/zopefoundation/zope.interface/pull/102>`_.
 
 
 4.4.3 (2017-09-22)
--- a/src/zope/interface/adapter.py
+++ b/src/zope/interface/adapter.py
@@ -257,7 +257,7 @@ class BaseAdapterRegistry(object):
         if value is None:
             new = ()
         else:
-            new = tuple([v for v in old if v is not value])
+            new = tuple([v for v in old if v != value])
 
         if new == old:
             return


src/zope/interface/tests/test_adapter.py
```diff
@@ -234,6 +234,17 @@ def test_unsubscribe_with_value_not_None_miss(self):
         registry.subscribe([IB1], None, orig)
         registry.unsubscribe([IB1], None, nomatch) #doesn't raise
         self.assertEqual(len(registry._subscribers), 2)
+    
+    def _instance_method_notify_target(self):
+        pass
```
mgedmin: This is the line that is not being executed, which causes a test coverage regression.  You could drop a `# pragma: nocover` comment next to it, or use @jamadden's favorite trick of replacing the `pass` with a docstring, e.g. `"""example subscriber that is a bound method"""` or something.

jamadden: Or my other favorite trick, raising an exception if it's called. The best one here would be `self.fail("Example method, not intended to be called.")`

mgedmin:
The problem with using `is` is that this makes it difficult to unregister bound methods.

jamadden:
> The problem with using is is that this makes it difficult to unregister bound methods.

Sure, if you rely on using that expression every time (and thus creating the new object). If you first retrieve the utility and then unregister it, you don't have a problem. That seems like the more general case to me. Being able to create a new object and unregister based on equality seems like a special case; it may not even make sense to support that special case given persistent objects, and the potential danger of false-positives.

Now, I understand that #12 was not actually about utilities but about subscription adapters or handlers. It is harder to first retrieve a registered handler than it is a utility (but not impossible), so that option still exists. Keeping a reference to the registered object is also a possibility and probably easier. But as for being able to register bound methods as a handler: isn't that a bit of a foot-gun already? They definitely don't work in persistent registries since they can't be pickled:
```pycon
Python 2.7.13 (default, May  5 2017, 09:16:47)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> class Foo(object):
...     def thing(self): pass
...
>>> f = Foo()
>>> t = f.thing
>>> t
<bound method Foo.thing of <__main__.Foo object at 0x101e82dd0>>
>>> import cPickle
>>> cPickle.dumps(t)
  File "/lib/python2.7/copy_reg.py", line 70, in _reduce_ex
TypeError: can't pickle instancemethod objects
>>> cPickle.dumps(t, 2)
cPickle.PicklingError: Can't pickle <type 'instancemethod'>: attribute lookup __builtin__.instancemethod failed
```

If the use for bound-methods-as-listeners is for test cases and `zope.component`s global registery, then normal test tearDown should already be cleaning them up.

I'm not really strongly opposed to this and won't object to seeing it merged (the main benefit for persistent objects would require another change),  I'm just struggling to understand the reasons to support it.

mgedmin:
I stumbled upon this in 2010 and I cannot for the life of me remember why I was trying to register/unregister bound methods.

I'm okay with saying 'don't do that', as long as the underlying zope.interface and the wrapper zope.component methods are in agreement (which they currently are not).

dwt:
> An advantage of is over ==, aside from speed is that is doesn't unghost persistent objects. In some cases, unghosting them may not even be possible (POSKeyError) so is can be very valuable.

I think the point here is that `{un}registerHandler` checks if the handler is registered with `==` but then `unregister` uses `is` to find the actual handler - this disagreement causes `unregisterHandler` to return true (and trigger unghosting), while the actual handler is not unregistered. This is the bug that this pull request tries to fix.

We found that it makes it quite a bit easier to register (and unregister) handlers in unit tests that don't pull in anything besides `unittest.TestCase`. Also please keep in mind that `zope.interface` is used widely outside the zope community.

src/zope/interface/tests/test_adapter.py
dwt: [x] done

icemac:
@dwt Is there still something to be done to get this PR ready for merge?

PR #103, opened 2017-09-19
Fix process of reference count during GC

y-fujisaki2:
call PyObject_GC_UnTrack() in tp_dealloc()
see the following sites for details:
 * https://bugs.python.org/issue31095
 * https://github.com/python/cpython/pull/2974

### DIFF
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -759,7 +759,7 @@ lookup_traverse(lookup *self, visitproc visit, void *arg)
 
 static int
 lookup_clear(lookup *self)
-{	
+{
   Py_CLEAR(self->_cache);
   Py_CLEAR(self->_mcache);
   Py_CLEAR(self->_scache);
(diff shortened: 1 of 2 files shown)


src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -759,7 +759,7 @@ lookup_traverse(lookup *self, visitproc visit, void *arg)
 
 static int
 lookup_clear(lookup *self)
-{
+{
```
mgedmin: I would prefer not to see unnecessary trailing whitespace being added to an unrelated function.

y-fujisaki2: I appreciate you removing it.

jamadden: @y-fujisaki2 I believe @mgedmin was requesting that you update the PR to avoid this problem. I am making the same request.

mgedmin:
Thank you!

mgedmin:
I've released zope.interface 4.4.3 to PyPI with this fix.

y-fujisaki2:
Thank you so much for follows!

PR #104, opened 2017-09-26
fix catching ImportError exception

j00zek:
The change avoids error when imported pkg_resources module does not contain declare_namespace attribute.

### DIFF
--- a/src/zope/__init__.py
+++ b/src/zope/__init__.py
@@ -1,7 +1,7 @@
 # this is a namespace package
 try:
-    import pkg_resources
-    pkg_resources.declare_namespace(__name__)
+    from pkg_resources import declare_namespace
+    declare_namespace(__name__)
 except ImportError:
     import pkgutil
     __path__ = pkgutil.extend_path(__path__, __name__)


mgedmin:
Curious: where did you find a `pkg_resources` without a `declare_namespace()`?

src/zope/__init__.py
```diff
@@ -1,7 +1,7 @@
 # this is a namespace package
 try:
-    import pkg_resources
-    pkg_resources.declare_namespace(__name__)
+    from pkg_resources import declare_namespace
+    declare_namespace(__name__)
 except ImportError:
     import pkgutil
     __path__ = pkgutil.extend_path(__path__, __name__)
```
mgedmin: So, AFAIU the `zope/__init__.py` is supposed to be identical in all `zope.*` packages (because on Python 2 you have no control which one ends up being installed), so if this one is wrong, we'll have to do a mass-update.

j00zek:
The issue exists on Kodi working on some Android distributions.

mgedmin:
`sha256sum */src/zope/__init__.py|sort|uniq -c -w 64` shows me that we currently have 9 different versions of the `__init__.py` file:

```
      1 0522c70ab9a4af12a9e670ff32ed49819950dab2ac1c6ca3941526bde134c24c  zope.wineggbuilder/src/zope/__init__.py
      4 108a1767c4f88fd0b07c46a0387a2d2c2a854cd0bfb24e83c6bd749004bf9b2a  zope.app.container/src/zope/__init__.py
     43 1be778dfcbcfc7b9bf92ce5ef574ed337bbb2c344ee5d492ba489bba45a6c8f3  zope.app.localpermission/src/zope/__init__.py
     56 6ec6215146e4fbd174909f206fef85ac72ccce7ef16888616f615bdb9c7782ba  zope.annotation/src/zope/__init__.py
      1 7c959323b8d7818ed60c0c9df572333bb8e1d7e5de972594ef6fb21c36cf5251  zope.paste/src/zope/__init__.py
      1 9fa352c3c08e8ce2ad4850a77bf3aa7ce5bf3f8c2ca63f1e65ff515bc07cea22  zope.login/src/zope/__init__.py
      1 a189cdc23980500d04d91c5582ddfa425b1eac44faf5bf3037a1835406b8bac3  zope.sqlalchemy/src/zope/__init__.py
     14 adccdf32427d5a19a9a66158e7d7bbf3bce9c5bccfed182651e8f2be97d48ae1  zope.apidoc/src/zope/__init__.py
      7 c32f8a09df53e7fcd81618f39169ca42c1f127190708c54179633d612447d3f5  zope.app.appsetup/src/zope/__init__.py
```

So I guess having different versions isn't too bad, as long as they're roughly equivalent.

@j00zek, do you know if Kodi uses any `zope.*` packages other than zope.interface?  How are users installing it?  (IIRC pip and easy_install do different things with namespace packages.)

jamadden:
Many or most of the `__init__.py` files do not have an `except ImportError` at all any more. When we have a direct dependency on `setuptools` (like we do here) we should be able to know that the import will work. The `except` block is untested code.

If the import fails, the environment is broken because setuptools and pkg_resources are broken. 

Maybe the environment that was causing problems has an ancient version of setuptools (it has changed a lot lately). We do not have any kind of version requirement on it. Maybe adding one will ensure that a new enough version is in place that this environment doesn't fail.

j00zek:
zope packages seems to be identical to yours to my knowledge.
I beleve they use https://github.com/mablae/script.module.zope.interface fork to build "script.module.zope.interface" addon, which is available in Kodi.
It can be automatically installed when any addon has it in its dependencies e.g. 
    <requires>
        <import addon="xbmc.python" version="2.25.0"/>
        <import addon="script.module.simplejson"    version="3.3.0"/>
        <import addon="script.module.twisted" version="13.2.1"/> 
    </requires>

Please note: I've found this issue on Android only. Same version of Kodi installed either or Windows or Linux works properly.

tseaver:
@j00zek What version of `setuptools` is installed in the environment where you detect the `ImportError`.  I would rather that we just pin a sufficiently-recent version of `setuptools` to avoid the change in this PR.

PR #108, opened 2017-12-02
Let mac 3.6 float in Travis

Rotonen:
3.6 should not get pinned to 3.6.0 on mac, but float with the 3.6 releases.

Also noticed Terryfy had broken upstream for how it is used here. They've fixed that now.

https://github.com/matthew-brett/multibuild/issues/86
https://github.com/MacPython/terryfy/issues/15
https://github.com/MacPython/terryfy/commit/bd0dd2c1cfcb285baf9fafb260a4ed47804c183f


Rotonen:
This one?

https://bugs.python.org/issue29943

If so, 3.6.0'd be the broken one and that is thus fixed upstream?

jamadden:
3.6.*1* was the broken release, as I understand it. It wasn't clear to me if the fix was in 3.6.2 for sure or got pushed back to 3.6.3...I didn't wade through everything in the NEWS file yet.

Rotonen:
The API change side of 29943 landed in the changelog as https://bugs.python.org/issue27867 (in 3.6.2 RC 1 in both ) if that helps you cross-check my postulation.

https://docs.python.org/3.6/whatsnew/changelog.html

icemac:
Is there a consensus that the fix is fine and can be merged?

jamadden:
I'm a slight -0, only because we *know* that there are versions of 3.6 that produce broken wheels, so it seems prudent to pin to a version that we *know* works. I can't tell by looking at the file what version we're building with, and other than doing a build, it's annoying to try to track that down. 

But it probably doesn't matter that much.

icemac:
Maybe we should try to use a more recent Py36 version for now.

Rotonen:
Should I resurrect this one or is the current state fine as it is?

icemac:
Python 3.7 is also pinned to 3.7.0. I'm not sure how to proceed.

Rotonen:
I'd let everything float. I've not seen any ABI drama in regards to 3.7.

icemac:
@jamadden Did you change your mind? You were the one who gave a `-0` on this PR.

jamadden:
That was in the past, and I haven't heard of any recent ABI issues, so I withdraw any objection I had raised. 😄

icemac:
Nice, then let's use a floating version for 3.7, too.

icemac:
Do we need an entry in the change log, too?

mgedmin:
Given that https://github.com/nedbat/coveragepy/issues/703 now breaks our CI builds, I think it's time to dust off this PR and merge it.

mgedmin:
#152 was merged instead of this one.  Shall we close it, or do we want to make 3.6 and 3.7 float?

(It may be a good idea to add 3.8 builds on MacOS as well.)

Rotonen:
I've not bothered with the paperwork since the new contributor agreement, so on a short to medium term no effort from me should be expected.

mgedmin:
(I don't think a change this simple is copyrightable and therefore it doesn't need paperwork, but eh.)

Rotonen:
> (I don't think a change this simple is copyrightable and therefore it doesn't need paperwork, but eh.)

But the Github and Travis machinery prevents me from working on PRs as I'm now not in the correct user groups for automation to kick in and do its thing. At least it did in October when I was going to do a quick fix for something and simply dropped it as did not want to figure out the paperwork.

PR #110, opened 2017-12-02
Drop testing and pypi-declared support of Python 3.3

Rotonen:
3.3 is already EOL and testing for it mostly just slows things down and gets in the way at this point.


tseaver:
ISTM that as long as we bump the version correctly, we can let folks who need 3.3 support pin to the last release (they likely have similar problems elsewhere, too).

We could add a [`python_requires` declaration](https://packaging.python.org/tutorials/distributing-packages/#python-requires) to help enforce this.

jamadden:
I would be +1 for the `python_requires` metadata. That way they won't unexpectedly get broken if they currently have a `>=` pin.

Rotonen:
So how to approach the release cutting?

Just bump it to 4.5 and also include https://github.com/zopefoundation/zope.interface/pull/102 or first cut 4.4.4 with that and then cut 4.5 with just dropping 3.3 support and enforcing that via setuptools?

Rotonen:
Bumped to 4.5.0 and can amend the PR if the change originally pending for 4.4.4 should get a release first.

tseaver:
I'd say leave the master branch targeting `4.5`:  if the change is desired, it will be trivial to backport the changes from #102 to create a `4.4.4`.

jamadden:
I have one remaining quibble, and that’s the “<3.7” in python_requires. Is that really necessary? That is, do we *know* this project won’t work on 3.7, to the point you shouldn’t even install it? 

I would prefer to avoid that restriction for a few reasons. I don’t like adding Yet Another thing that has to be updated for every python release (tox, Travis, appveyor, classifiers, changenote).  It’s going to cause serious pain on day one of general availability of 3.7 for every project that hasn’t already rereleased if that becomes widespread. It took many months to get the projects that make up the ZTK to have releases claiming 3.6 support, and very few, if any, had any actual code changes that were needed to make that happen. 

tl;dr I feel like pining an upper bound in python_requires is more hassle than benefit, hurts early adopters, and violates the spirit of “consenting adults”

Rotonen:
@jamadden thank you, amended.

Rotonen:
Still waiting on final approvals from the seconders of the review before merging.

hugovk:
There's also some `cp33` in .manylinux-install.sh and `py33` (and `py26`) in docs/hacking.rst that can go.

Also does the `!=3.3` format work for `python_requires`? Most examples I've seen are like `!=3.3.*`:

https://stackoverflow.com/a/44660547/724176

mgedmin:
Other zopefoundation repos also use `3.X.*`:

```
mg@platonas: ~/src/zopefoundation $ grep python_requires */setup.py
z3c.formui/setup.py:    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
ZEO/setup.py:      python_requires='>=2.7.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
ZODB/setup.py:    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
zope.hookable/setup.py:      python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
zope.password/setup.py:      python_requires='>=2.7.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
```

Rotonen:
Also dropped 3.3 from `.manylinux-install.sh` and `tox.ini`.

I'll still go through `docs/hacking.rst` as that also seems quite out of date.

PR #112, opened 2017-12-07
Use LuaLaTeX for PDF docs

### LINKED ISSUE #111: Builds on RTD broken
According to https://readthedocs.org/projects/zopeinterface/builds/ the builds on RTD are broken for a while. So no changes will show up there.

mgedmin:
See http://www.sphinx-doc.org/en/stable/config.html#confval-latex_engine

Ought to fix #111, provided that ReadTheDocs supports LuaLaTeX.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -188,6 +188,10 @@ latex_elements = {
 #'preamble': '',
 }
 
+# pdflatex can't handle Cyrillic out of the box, but xetext/lualatex should be
+# able to cope
+latex_engine = 'lualatex'
+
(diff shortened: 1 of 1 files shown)


mgedmin:
This fixes `make latexpdf` for me on Ubuntu 17.10 (after I also apt installed `texlive-luatex`).  

[ReadTheDocs say they have `texlive-full` installed](http://docs.readthedocs.io/en/latest/builds.html#packages-installed-in-the-build-environment), and I think it's reasonable to assume that that pulls in `texlive-luatex`.

tseaver:
LGTM.

PR #115, opened 2018-03-08
Synchronize and simplify zope/__init__.py

### LINKED ISSUE #114: Sync namespace packages approach
Hello.

As mentioned in Python docs [0]:
> **Every** distribution that uses the namespace package must include an identical `__init__.py`. If any distribution does not, it will cause the namespace logic to fail and the other sub-packages will not be importable. Any additional code in `__init__.py` will be inaccessible.

but it seems that zope.deprecation [1] and zope.configuration [2] have different content in namespace `__init__.py` files than zope.interface.

It would be nice to sync namespace `__init__.py` files across modules maintained by Zope foundation so they will be installable in parallel.

[0] https://packaging.python.org/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages
[1] https://github.com/zopefoundation/zope.deprecation/blob/master/src/zope/__init__.py
[2] https://github.com/zopefoundation/zope.configuration/blob/master/src/zope/__init__.py

Possibly related to:
* https://github.com/zopefoundation/zope.interface/issues/107
* https://github.com/zopefoundation/zope.interface/issues/68
* https://github.com/zopefoundation/zope.interface/issues/97

jamadden:
We *do* have a dependency on setuptools, so the try/except block should be unnecessary. At worst it might hide installation errors.

Fixes #114.

### DIFF
--- a/src/zope/__init__.py
+++ b/src/zope/__init__.py
@@ -1,7 +1,1 @@
-# this is a namespace package
-try:
-    import pkg_resources
-    pkg_resources.declare_namespace(__name__)
-except ImportError:
-    import pkgutil
-    __path__ = pkgutil.extend_path(__path__, __name__)
+__import__('pkg_resources').declare_namespace(__name__)


jamadden:
Thanks!

PR #120, opened 2018-07-19
Publish the documentation for zope.interface.common

jamadden:
I was working on documenting another project and wanted to link to them (especially zope.interface.common.idatetime) but discovered they weren't published.

This makes the API page even longer and more unwieldy. I'm working on a followup PR that splits it into more manageable chunks.


jamadden:
Thank you!

PR #127, opened 2018-10-23
Fix 'verifyObject' for class objects with staticmethods on Python 3.

### LINKED ISSUE #126: BrokenMethodImplementation "doesn't allow enough arguments" error when using staticmethods on Python 3
Verifying a class using `@provider` behaves differently on Python 2.7 vs. Python 3.7 with zope.interface 4.5.0:

```python
import zope.interface
from zope.interface.verify import verifyObject

class IFoo(zope.interface.Interface):
    def bar(a, b):
        pass

@zope.interface.provider(IFoo)
class Foo(object):
    @staticmethod
    def bar(a, b):
        pass

assert verifyObject(IFoo, Foo)
```

On Python 2.7, this code passes. On Python 3.7 it raises a `BrokenMethodImplementation`:

```pytb
  File "/Users/andy/test/.tox/py37/lib/python3.7/site-packages/zope/interface/verify.py", line 105, in verifyObject
    return _verify(iface, candidate, tentative, vtype='o')
  File "/Users/andy/test/.tox/py37/lib/python3.7/site-packages/zope/interface/verify.py", line 97, in _verify
    raise BrokenMethodImplementation(name, mess)
zope.interface.exceptions.BrokenMethodImplementation: The implementation of bar violates its contract
        because implementation doesn't allow enough arguments.
```

jamadden:
Fixes #126

### DIFF
--- a/src/zope/interface/tests/test_verify.py
+++ b/src/zope/interface/tests/test_verify.py
@@ -557,5 +557,26 @@ class Test_verifyObject(Test_verifyClass):
         self.assertRaises(DoesNotImplement,
                           self._callFUT, IDummyModule, dummy)
 
+    def test_staticmethod_hit_on_class(self):
+        from zope.interface import Interface
+        from zope.interface import provider
+        from zope.interface.verify import verifyObject
+
+        class IFoo(Interface):
+
+            def bar(a, b):
+                pass
+
+        @provider(IFoo)
+        class Foo(object):
+
+            @staticmethod
+            def bar(a, b):
+                pass
+
+        # Don't use self._callFUT, we don't want to instantiate the
+        # class.
+        verifyObject(IFoo, Foo)
+
 class OldSkool:
     pass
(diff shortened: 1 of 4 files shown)


src/zope/interface/tests/test_verify.py
```diff
@@ -557,5 +557,26 @@ class IDummyModule(Interface):
...
 
+    def test_staticmethod_hit_on_class(self):
+        from zope.interface import Interface
+        from zope.interface import provider
+        from zope.interface.verify import verifyObject
+
+        class IFoo(Interface):
+
+            def bar(a, b):
+                pass
```
rouge8: I think to make coverage happy you can change this to `raise NotImplementedError()`

jamadden: Whoops, usually I'm pretty good about that kind of thing. Thanks!

jamadden:
Thanks! 4.6.0 has been released to PyPI.

rouge8:
Thanks for the fast fix!

PR #139, opened 2019-04-18
Fix issue with missing `__init__.py`.

### LINKED ISSUE #68: Missing __init__.py warning when importing pkg_resources
Hi. I found that a very simple script can emit an annoying warning about zope.interface. Type this in an empty directory:
``` bash
python2 -m virtualenv env
./env/bin/pip install zope.interface
./env/bin/python -Werror -c "import pkg_resources"
```
The warning produced is:
```
  File "<string>", line 1, in <module>
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3019, in <module>
    @_call_aside
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3003, in _call_aside
    f(*args, **kwargs)
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3047, in _initialize_master_working_set
    for dist in working_set
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3047, in <genexpr>
    for dist in working_set
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2577, in activate
    declare_namespace(pkg)
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2151, in declare_namespace
    _handle_ns(packageName, path_item)
  File "./env/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2071, in _handle_ns
    loader = importer.find_module(packageName)
  File "/usr/lib64/python2.7/pkgutil.py", line 186, in find_module
    file, filename, etc = imp.find_module(subname, path)
ImportWarning: Not importing directory './env/lib/python2.7/site-packages/zope': missing __init__.py
```
It is still effective with the current github code `./env/bin/pip install git+https://github.com/zopefoundation/zope.interface.git `

It is a very annoying message since on my django project, it represents 90% of the warnings emitted when I launch my unit tests.

I have tested this with python 2.7.12 and python 3.5.2.

icemac:
Fixes #68.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changes
 4.6.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix issue with missing ``__init__.py`` by requiring a ``setuptools`` version
+  which supports PEP 420.
+  (`#68 <https://github.com/zopefoundation/zope.interface/issues/68>`_)
 
 
 4.6.0 (2018-10-23)
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,7 @@ setup(name='zope.interface',
       include_package_data=True,
       zip_safe=False,
       tests_require=tests_require,
-      install_requires=['setuptools'],
+      install_requires=['setuptools >= 40.3'],
       python_requires=', '.join((
           '>=2.7',
           '!=3.0.*',


jamadden:
I would suggest we use caution pinning the runtime version of setuptools to any specific version; it's a very fundamental library and so is this one. I think I recall complaints in gevent when I started using and requiring newer setuptools features; apparently some environments have a difficult time changing setuptools versions.

icemac:
@mgedmin The problem only seems to occur in certain environments but there for every namespace package I fear.

icemac:
I am closing this PR as the required `setuptools` version might be too new for some users not having the problem. For now it should be enough to have an open issue redirecting to the proper solution without enforcing it.

d-maurer:
Michael Howitz wrote at 2019-5-15 05:59 +0000:
>@mgedmin The problem only seems to occur in certain environments but there for every namespace package I fear.

I got the impression that Python 3 no longer needs an `__init__.py`
in a package. More precisely: if there is one, then the package is
interpreted as a "normal" package, otherwise as a namespace package.

In addition, it seems to me, that modern `pip` (maybe also `setuptools`)
versions behave similarly for Python 2, too.


Maybe related: I have observed strange problems when
mixing "pip" installed and "buildout" installed (namespace) packages.
My (still incomplete) analysis seems to hint toward a Python bug:
a `__path__` which should auto-update after a change to `sys.path`
remained un-updated.

jugmac00:
> I got the impression that Python 3 no longer needs an `__init__.py` in a package.
There was a change with Python 3.3
https://www.python.org/dev/peps/pep-0420/

PR #142, opened 2019-08-17
Port the documentation examples to Python 3

### LINKED ISSUE #119: Documentation on implements is out of date
The current documentation still shows `zope.interface.implements` as the method for declaring that a class implements an interface. Best I can tell, this will always throw an error in python 3.6

mgedmin:
Fixes #119.

I don't see where Travis CI is running `sphinx-build -b doctest`.  If it doesn't, that seems like a bad oversight (that I don't currently have energy to rectify.  Help, anyone?)

I haven't read the surrounding text closely enough to be sure it still makes sense after my changes.  A close review would be appreciated.

I'm not sure what to do with the code example for `classImplements` that I had to rewrite to use `@implementer` because `classImplements` does not work any more.

If Zope Foundation had any money maybe they could hire a technical writer to clean this up....  (RIP Zope Foundation, long live Plone Foundation!)

### DIFF
--- a/docs/api/declarations.rst
+++ b/docs/api/declarations.rst
@@ -61,7 +61,7 @@ Consider the following example:
 .. doctest::
 
    >>> from zope.interface import implementedBy
-   >>> from zope.interface import implements
+   >>> from zope.interface import implementer
    >>> from zope.interface import classImplementsOnly
    >>> from zope.interface import Interface
    >>> class I1(Interface): pass
@@ -72,10 +72,12 @@ Consider the following example:
    ...
    >>> class I4(Interface): pass
    ...
-   >>> class A(object):
-   ...   implements(I3)
-   >>> class B(object):
-   ...   implements(I4)
+   >>> @implementer(I3)
+   ... class A(object):
+   ...     pass
+   >>> @implementer(I4)
+   ... class B(object):
+   ...     pass
    >>> class C(A, B):
    ...   pass
    >>> classImplementsOnly(C, I1, I2)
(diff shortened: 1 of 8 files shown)


mgedmin:
See https://github.com/zopefoundation/zope.interface/pull/143#issuecomment-522329065 for some instructive examples about the reliability of coveralls dot io.

AFAICT sphinx doctests were never included in the coverage measurements on Travis so changes to  docs/*.rst should not have had any effect on coverage numbers.

mgedmin:
Well, this did not make the 4.7.0 release, unfortunately.  (I forgot about it.)

docs/api/declarations.rst
```diff
@@ -324,6 +335,8 @@ Removing an interface that is provided through the class is not possible:
 classProvides
 -------------
 
+(The `provider` decorator is preferred to this.)
```
jamadden: I think the example here should be moved down to the `provider` section. 

This section should either be removed, or a comment added to say that `classProvides` is deprecated and only works on Python 2; no example is provided.

mgedmin: Good idea, thank you, done.

docs/api/declarations.rst
```diff
@@ -37,15 +37,15 @@ implementer_only
 implements
 ----------
 
-(The `implementer` decorator is preferred to this.)
+(Does not work on Python 3.  Use the `implementer` decorator instead.)
```
jamadden: > (These notes should probably use some Sphinx directive to stand out
better, but I'm not well versed in Sphinx and I don't know which one I
should use.)

`.. caution:: Does not work on Python 3.  Use the `implementer` decorator instead.` is one way. 

(Themes usually style `.. note::` and `.. caution::`. They may or may not style `.. warning::`. They usually don't style `.. important::` and `.. tip::` in my experience.)

PR #144, opened 2019-10-09
provides taggedvalue inheritance

mamico:
provides query/get/list taggedvalue with subclass inheritance
fix: https://github.com/plone/plone.dexterity/issues/57 and similar issues

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ Changes
 
 - Drop support for Python 3.4.
 
+- Fix ``queryTaggedValue``, ``getTaggedValue``, ``getTaggedValueTags``
+  subclass inheritance. See `PR 144
+  <https://github.com/zopefoundation/zope.interface/pull/144>`_.
 
 4.6.0 (2018-10-23)
 ------------------
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -452,6 +452,31 @@ class InterfaceClass(Element, InterfaceBase, Specification):
         if errors:
             raise Invalid(errors)
 
+    def queryTaggedValue(self, tag, default=None):
+        """ Returns the value associated with 'tag'. """
+        value = Element.queryTaggedValue(self, tag, default=_marker)
+        if value is not _marker:
+            return value
+        for base in self.__bases__:
+            value = base.queryTaggedValue(tag, default=_marker)
+            if value is not _marker:
+                return value
+        return default
+
+    def getTaggedValue(self, tag):
+        """ Returns the value associated with 'tag'. """
+        value = self.queryTaggedValue(tag, default=_marker)
+        if value is _marker:
+            raise KeyError(tag)
+        return value
+
+    def getTaggedValueTags(self):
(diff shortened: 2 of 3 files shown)


mamico:
> LGTMish, 

good, thanks.

> but I don't see
>  * a changelog update

done.

>  * documentation update
>
> Also, is this a backwards-incompatible change (requiring a major version bump), a new feature (requiring a minor version bump), or a bugfix?

I ask for your support on that. IMHO it's a bugfix, no changes. The only backwards-incompatible could be if someone expects the current _wrong_ behavior, where tag on a class don't inherits in a subclass...

icemac:
Coveralls seems to be broken in itself: it lists all files several times.

mgedmin:
> I think of this change not as a simple bugfix more as a missing feature as you only had to write new code instead of changing existing one (besides in the tests.)

There's a difference between overriding an inherited method (to fix behaviour) and adding a new method (to add behaviour), but merely looking at the diff is not enough to distinguish these two cases (which was why I asked).

icemac:
@mgedmin Right, I did not look deep enough into the code.

I know and use the tagging features sporadically. The issue which is fixed here has never been a problem for me.
I am not sure if this PR breaks someone's code so let's make the screaming test.

mamico:
@icemac @mgedmin thanks a lot for the reviews. 
This i not an urgent bug, in my opinion, but a bit annoying on cases like https://github.com/plone/plone.dexterity/issues/57 filled by @volkerjaenisch and @jensens.
I ran into the same problem and wasted time to know the cause.
BTW plone dexterity uses extensively schema tagging for field permission, sorting, ...

jensens:
This missing inheritance of tagged values always felt like a bug to me, even if one may argue its a missing feature. It is just not what one would expect. Also, it makes life harder than it should.

:+1: for merging this.

mamico:
> @mamico Do you need a release of `zope.interface` to use this feature somewhere?

@icemac  as I said above I've workaround. No particular urgency for me. Thanks.

jamadden:
FWIW, this turned out to be a breaking change for us. We apparently had code (in [nti.externalization](https://github.com/NextThought/nti.externalization)) that implicitly relied on tagged values *not* being inherited. There's code that examines classes in a module, and if they implement an interface that provides a particular tagged value, a utility is registered with a name from the value of that tag. With 4.7.0, we started registering duplicate utilities because we started finding classes that only implemented a derived interface. Simplified example:

```python
class IPublic(Interface):
    taggedValue('tagged-value', 'public')

class IPrivate(IExternal):
    pass

@implementer(IExternal)
class Public(object): pass

@implementer(IPrivate)
class PrivateImplementation(object): pass
```

With 4.6, we'd register `Public` with a name of `public`. With 4.7, we're trying to register `Public` with a name of `public` and `PrivateImplementation` also with a name of `public`. This leads to `zope.configuration.config.ConfigurationConflictError: Conflicting configuration actions`.

I think I can workaround this by calling `Element.queryTaggedValue(iface, ...)` instead of `iface.queryTaggedValue(...)`...

PR #145, opened 2019-11-11
Add support for Python 3.8

cjwatson:


.travis.yml
```diff
@@ -12,6 +12,10 @@ matrix:
           python: 3.7
           dist: xenial
           sudo: true
+        - os: linux
+          python: 3.8
+          dist: xenial
```
jugmac00: You can leave out `dist: xenial`.

.travis.yml
```diff
@@ -12,6 +12,10 @@ matrix:
...
+          sudo: true
```
jugmac00: The `sudo` keyword had been deprecated quite a while ago - it does nothing any more and should be removed.

.travis.yml
cjwatson: Can I remove it from the 3.7 case too?  I was just copying that.

mgedmin: It's fine, I've plans of updating the travis.yml in alignment with the other packages I've updated for https://github.com/zopefoundation/meta/issues/11 some time this week anyway.

.travis.yml
```diff
@@ -31,6 +35,9 @@ matrix:
         - os: osx
           language: generic
           env: TERRYFY_PYTHON='macpython 3.7.0'
+        - os: osx
+          language: generic
+          env: TERRYFY_PYTHON='macpython 3.8.0'
```
mgedmin: Oh, is 3.8 available from terryfy already!  That's nice!

CHANGES.rst
```diff
@@ -10,6 +10,8 @@ Changes
   subclass inheritance. See `PR 144
   <https://github.com/zopefoundation/zope.interface/pull/144>`_.
 
+- Add support for Python 3.8.
+
```
mgedmin: I like two blank lines between ReST sections, but we haven't been consistent with this in the past.

.travis.yml
mgedmin: Yes, I've been removing both `dist: xenial` and `sudo: true` as I upgrade the travis ymls.

.travis.yml
mgedmin: [Apparently not](https://travis-ci.org/zopefoundation/zope.interface/jobs/610282558?utm_medium=notification&utm_source=github_status).

.travis.yml
jugmac00: Yes, it is no longer necessary as xenial now is the default environment.
https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

.travis.yml
mgedmin: I'm not a Mac user and know little about these things, but https://github.com/MacPython/terryfy looks nearly abandoned: last commit in 2018, no issues requesting Python 3.8 support, a README that directs users to https://github.com/matthew-brett/multibuild which "gets more development attention".

There's also https://github.com/joerick/cibuildwheel.

cjwatson: OK, I know nothing about Mac builds.  How about I just drop this for now, and somebody with more clue can set it up?

cjwatson: I've done that now.

mgedmin: That was always my approach too: ignore 3.8 on Mac until some Mac user steps up. ;)

mgedmin:
I wouldn't pay too much attention to coveralls numbers, it seems they're not aggregating their data properly.

PR #146, opened 2019-11-11
Always build manylinux wheels

mgedmin:
This PR makes it so we _always_ build manylinux wheels, but only upload
them for git tags.

Highlights:

- Move regular Linux Python builds outside the build matrix, for
  conciseness.

- Switch manylinux1 builds to manylinux2010 (they should still get the
  manylinux1 tag if they're compatible, but they'll be built with a more
  modern compiler toolchain).

- Move twine upload from .manylinux.sh into .travis.yml's after_success,
  right next to the twine upload used for Mac OS builds.

- This means we can't short-circuit the builds with 'exit 0' in
  before_install:, so overwrite the install: and script: sections too.

- Move the 'docker pull' from .manylinux.sh into the install: section
  so it's not empty.

- Provide twine credentials via environment variables instead of
  command-line/pypirc file.

- Put related commands inside a single multiline if statement.

- Use setup.py -q test (quiet build) instead of setup.py test -q (which
  does nothing at all AFAICT).

This mirrors changes made to other zopefoundation packages tracked in
https://github.com/zopefoundation/meta/issues/11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,78 +1,94 @@
 language: python
-sudo: false
-matrix:
-    include:
-        - os: linux
-          python: 2.7
-        - os: linux
-          python: 3.5
-        - os: linux
-          python: 3.6
-        - os: linux
-          python: 3.7
-        - os: linux
-          python: 3.8
-        - os: linux
-          python: pypy
-        - os: linux
-          python: pypy3
-        # It's important to use 'macpython' builds to get the least
-        # restrictive wheel tag. It's also important to avoid
-        # 'homebrew 3' because it floats instead of being a specific version.
-        - os: osx
-          language: generic
-          env: TERRYFY_PYTHON='macpython 2.7'
-        - os: osx
-          language: generic
-          env: TERRYFY_PYTHON='macpython 3.5'
-        - os: osx
-          language: generic
-          env: TERRYFY_PYTHON='macpython 3.6.0'
-        - os: osx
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,78 +1,94 @@
...
-    - echo password=$PYPIPASSWORD                      >> ~/.pypirc
-    - if [[ $TRAVIS_TAG && "$TRAVIS_OS_NAME" == "osx" ]]; then pip install twine; fi
-    - if [[ $TRAVIS_TAG && "$TRAVIS_OS_NAME" == "osx" ]]; then python setup.py bdist_wheel; fi
-    - if [[ $TRAVIS_TAG && "$TRAVIS_OS_NAME" == "osx" ]]; then twine upload dist/*; fi
+  - coveralls
+  - |
+    if [[ $TRAVIS_TAG && "$TRAVIS_OS_NAME" == "osx" ]]; then
+      # macpython 3.5 doesn't support recent TLS protocols which causes twine
+      # upload to fail, so we use the system Python to run twine
+      /usr/bin/python -m pip install twine
```
mgedmin: Oops I messed this up: https://travis-ci.org/zopefoundation/zope.interface/builds/610352389

```
/usr/bin/python: No module named pip
```

mgedmin: → https://github.com/zopefoundation/zope.interface/pull/148

PR #147, opened 2019-11-11
Clean up appveyor.yml

mgedmin:
- Make sure that the Python we want actually exists (and install it if
  necessary) -- this will make our lives easier next year, when Python
  3.9 shows up

- Make the test run quiet (setup.py test -q ignores the -q)

- Pass --skip-existing to twine upload so the build of the tag can be
  retried if necessary without causing spurious failures

I think it might also be a good idea to add a custom version template to
avoid nonsense like https://ci.appveyor.com/project/mgedmin/zope-interface/builds/28770734,
but I'm not entirely sure what happened there and whether a custom
version template would help.  (All I know is all my project appveyors
YML have `version: build-{build}-{branch}` and I've never seen that kind
of error.)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,6 +18,13 @@ environment:
 
 install:
   - "SET PATH=C:\\Python%PYTHON%;c:\\Python%PYTHON%\\scripts;%PATH%"
+  - ps: |
+      $env:PYTHON = "C:\\Python${env:PYTHON}"
+      if (-not (Test-Path $env:PYTHON)) {
+        curl -o install_python.ps1 https://raw.githubusercontent.com/matthew-brett/multibuild/11a389d78892cf90addac8f69433d5e22bfa422a/install_python.ps1
+        .\install_python.ps1
+      }
+  - ps: if (-not (Test-Path $env:PYTHON)) { throw "No $env:PYTHON" }
   - echo "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /x64 > "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\bin\amd64\vcvars64.bat"
   - pip install -e .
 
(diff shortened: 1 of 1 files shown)


mgedmin:
> I don't understand the comment about a custom version template. For whatever reason, I can't see any test output at https://ci.appveyor.com/project/mgedmin/zope-interface/builds/28770734.

Appveyor assigns each build a version number: https://www.appveyor.com/docs/build-configuration/#build-versioning.  This number must be unique.

Somehow, when I ran zest.releaser and pushed to GitHub, appveyor managed to assign the same build number to two different builds:

- https://ci.appveyor.com/project/mgedmin/zope-interface/builds/28770731
- https://ci.appveyor.com/project/mgedmin/zope-interface/builds/28770734

as a result, one of the two builds failed with the error message you've screenshotted.  I've restarted the build for that commit, which created a new version number:

- https://ci.appveyor.com/project/mgedmin/zope-interface/builds/28771018

I don't want to do this every time somebody pushes a release.  I think it's a bug on appveyor's side, but I wonder if I could use the [configurable version format setting in appveyor.yml](https://www.appveyor.com/docs/build-configuration/#build-versioning) to avoid it in the future?
This might work if {branch} would expand to the git tag, rather than 'master' for one of the two builds, but I'm not sure it will.

mgedmin:
Looking at https://ci.appveyor.com/project/mgedmin/findimports/history I think it would work!

Ekrano nuotrauka iš 2019-11-11 17-58-25

There was no race condition in the screenshot ({build} was expanded to 34 and 35), but if that had happened, there would've been no conflict, as {branch} was expanded to the git tag for one build and 'master' for the other build.

appveyor.yml
```diff
@@ -26,13 +33,13 @@ build_script:
...
-  - python setup.py test -q
+  - python setup.py -q test
 
 artifacts:
   - path: 'dist\*.whl'
     name: wheel
 
 deploy_script:
-  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload dist/* }
+  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine --skip-existing upload dist/* }
```
mgedmin: This part was not so good: https://ci.appveyor.com/project/mgedmin/zope-interface/builds/28777006/job/g40lp9dhkof95mf2

Fixed in 1a84324.

PR #148, opened 2019-11-11
Fix Mac OS wheel uploads

mgedmin:
Hopefully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,7 @@ after_success:
     if [[ $TRAVIS_TAG && "$TRAVIS_OS_NAME" == "osx" ]]; then
       # macpython 3.5 doesn't support recent TLS protocols which causes twine
       # upload to fail, so we use the system Python to run twine
+      /usr/bin/python -m ensurepip
       /usr/bin/python -m pip install twine
       twine upload --skip-existing dist/*
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ install:
 script:
   - python --version
   - coverage run setup.py -q test
-  - python setup.py bdist_wheel
+  - python setup.py -q bdist_wheel
 
 after_success:
   - coveralls
@@ -78,7 +78,8 @@ after_success:
       # macpython 3.5 doesn't support recent TLS protocols which causes twine
       # upload to fail, so we use the system Python to run twine
       /usr/bin/python -m ensurepip --user
-      /usr/bin/python -m pip install --user -U pip twine
+      /usr/bin/python -m pip install --user -U pip
+      /usr/bin/python -m pip install --user -U twine
       /usr/bin/python -m twine check dist/*
       if [[ $TRAVIS_TAG ]]; then
         /usr/bin/python -m twine upload --skip-existing dist/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ after_success:
       # macpython 3.5 doesn't support recent TLS protocols which causes twine
       # upload to fail, so we use the system Python to run twine
       /usr/bin/python -m ensurepip --user
-      /usr/bin/python -m pip install --user twine
+      /usr/bin/python -m pip install --user -U pip twine
       /usr/bin/python -m twine check dist/*
       if [[ $TRAVIS_TAG ]]; then
         /usr/bin/python -m twine upload --skip-existing dist/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ after_success:
       # upload to fail, so we use the system Python to run twine
       /usr/bin/python -m ensurepip --user
       /usr/bin/python -m pip install --user -U pip
-      /usr/bin/python -m pip install --user -U twine
+      /usr/bin/python -m pip install --user -U -I twine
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -77,6 +77,7 @@ after_success:
     if [[ $TRAVIS_TAG && "$TRAVIS_OS_NAME" == "osx" ]]; then
       # macpython 3.5 doesn't support recent TLS protocols which causes twine
       # upload to fail, so we use the system Python to run twine
+      /usr/bin/python -m ensurepip
```
jamadden: I'm pretty sure [Travis doesn't run this script as root](https://travis-ci.org/zopefoundation/zope.interface/jobs/610352400#L86), so [all of the install commands](https://github.com/zopefoundation/BTrees/issues/113#issuecomment-552103774) need the `--user` flag. I'm not sure that the user directory is on the default path, so I think the rest of the commands need modified to reference the module:

```
$ /usr/bin/python -mensurepip --user
$ /usr/bin/python -m pip install -U --user pip
$ /usr/bin/python -m pip install -U --user twine
$ /usr/bin/python -m twine upload --skip-existing dist/*
```

mgedmin: (The reason I'm floundering is I hate pip install --user, no rational reason.)

.travis.yml
```diff
@@ -77,18 +77,20 @@ after_success:
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       # macpython 3.5 doesn't support recent TLS protocols which causes twine
       # upload to fail, so we use the system Python to run twine
-      virtualenv -p /usr/bin/python .twine-env && . .twine-env/bin/activate
+      /usr/bin/python3 -m venv .twine-env && . .twine-env/bin/activate
```
jamadden: Sorry, Apple only ships /usr/bin/python as a symlink to /usr/bin/python2.7.

mgedmin: Can the Travis documentation be any vaguer?  https://docs.travis-ci.com/user/reference/osx/#python-related-tools

jamadden: Agreed, it's pretty sparse.

I looked into using the homebrew addon to install Python, but it seems that homebrew [doesn't give control over the Python version](https://docs.brew.sh/Homebrew-and-Python) --- you get 2 or 3, and that's as fine as it goes.

However, if homebrew can install pyenv, then pyenv can install a specific Python version (I use pyenv to install specific Python versions for gevent).  One nice thing is that pyenv takes care to build Python using a functioning OpenSSL on macOS. I just don't know much about homebrew or the specific incantations that would be required here.

mgedmin:
The last version works, but now I'm having second thoughts:

- this makes twine use Python 2, which is going to be EOL in about seven weeks.
- I'm not sure _which_ minor version of Python 2 we're getting, or if it will be able to talk to PyPI

On the plus side, this is definitely less broken than what I've already merged in #146.

jamadden:
> this makes twine use Python 2, which is going to be EOL in about seven weeks.

It's Apple's system python, so as long as macOS 10.13 or 10.14 is supported by Apple, it will work. Also, we only need it for as long as we support Python 3.5, which itself goes EOL pretty soon.

> I'm not sure which minor version of Python 2 we're getting, or if it will be able to talk to PyPI

If we're on 10.13 or 10.14, we can know that it will be able to. We can explicitly control [which macOS version we use](https://docs.travis-ci.com/user/reference/osx/#macos-version). I don't know about 10.13, but on 10.14, /usr/bin/python is 2.7.16.

mgedmin:
> > this makes twine use Python 2, which is going to be EOL in about seven weeks.

> It's Apple's system python, so as long as macOS 10.13 or 10.14 is supported by Apple, it will work. 

Yes, but Twine or one of its dependencies might break 2.7 compatibility, and if it doesn't use python_requires in its setup.py, that would break us.

> Also, we only need it for as long as we support Python 3.5, which itself goes EOL pretty soon.

Not soon enough IMHO.  I'd love to drop 2.7 and 3.5 at the same time.  Oh well, this can't be helped.

I wish I could do this pip install --user dance only for 3.5, but I don't have the stamina for ten more inevitable "will it work this time" pushes.  I'll merge this if you approve it.

jamadden:
> Yes, but Twine or one of its dependencies might break 2.7 compatibility, and if it doesn't use python_requires in its setup.py, that would break us.

Fair point. Twine 2/master has already dropped support for everything except 3.6+ in order to use type annotations. Twine itself does [have a correct python_requires](https://github.com/pypa/twine/blob/master/setup.py#L68) for both master [and 1.x](https://github.com/pypa/twine/blob/maint/1.x/setup.py#L71) but I haven't walked through its dependency tree. 

>> Also, we only need it for as long as we support Python 3.5, which itself goes EOL pretty soon.
> Not soon enough IMHO.

Just as a data point, looking at https://pypistats.org/packages/zope-interface, 3.5 is never more than a distant fourth place for number of installs, less than 5%. macOS appears to have around half the install size of Windows. (These numbers are highly skewed by CI systems which tend to be Linux.)

mgedmin:
Good news: this works!  https://travis-ci.org/zopefoundation/zope.interface/builds/610447412 shows 2.7 and 3.5 wheel uploads have succeeded.  (The others are still in progress.)

PR #154, opened 2020-01-22
Use tp_members in SpecificationBase and ClassProvidesBase; stop using private undoc API

jamadden:
Instead of storing 1 and 2 attributes, respectively, in `__dict__`. This lets us eliminate inst_attr and its use of an undocumented API.

It also helps reign in the size of these objects. Before, on CPython 8, a ClassProvides object that subclasses both SpecificationBase and ClassProvidesBase required 408 bytes: 48 for the object, 360 for the dict.

Now it requires 72 + 232 = 304.

With a bit of work on defining `__slots__` in Python, that can be brought much lower still, I think.

zope.security tests pass with this installed.

The .c file was mostly spaces with a smattering of tabs thrown in; this PR also removes the tabs for consistency so `?w=1` might be helpful.

### DIFF
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -15,6 +15,12 @@
 #include "Python.h"
 #include "structmember.h"
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wmissing-field-initializers"
+#endif
+
 #define TYPE(O) ((PyTypeObject*)(O))
 #define OBJECT(O) ((PyObject*)(O))
 #define CLASSIC(O) ((PyClassObject*)(O))
@@ -32,7 +38,7 @@
 static PyObject *str__dict__, *str__implemented__, *strextends;
 static PyObject *BuiltinImplementationSpecifications, *str__provides__;
 static PyObject *str__class__, *str__providedBy__;
-static PyObject *empty, *fallback, *str_implied, *str_cls, *str_implements;
+static PyObject *empty, *fallback, *str_implements;
 static PyObject *str__conform__, *str_call_conform, *adapter_hooks;
 static PyObject *str_uncached_lookup, *str_uncached_lookupAll;
 static PyObject *str_uncached_subscriptions;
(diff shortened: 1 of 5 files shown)


src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -573,7 +611,8 @@ __adapt__(PyObject *self, PyObject *obj)
     {
       PyObject *implied;
 
-      implied = inst_attr(decl, str_implied);
+      //implied = ((Spec*)decl)->_implied;
+      implied = PyObject_GetAttrString(decl, "_implied");
```
mgedmin: ?

jamadden: Whoops. Sorry. I was checking to see if there were any tests in place that would detect the ref leak. Didn't mean to leave that in.

jamadden:
> lgtm

Thanks!

> but I don't know anything about the Python C API or writing extension modules ;)

Neither do I! That's what makes this ever so exciting 😉 

I know that adding `__slots__` to more places might change how things pickle and would need to be taken into account for compatibility. Moving attributes from `__dict__` to `tp_members` is the C equivalent of `__slots__` so I looked into pickling. Starting from the two base classes changed here, these are all the defined descendent classes:

- zope.interface.Provides
  Defines its own `__reduce__` so pickles the same everywhere.
- zope.interface.declarations.ClassProvides
   Defines its own `__reduce__` so pickles the same everywhere.
- zope.interface.declarations.Implements
  Defines its own `__reduce__` so pickles the same everywhere.
- zope.interface.interface.InterfaceClass
  Defines its own `__reduce__` so pickles the same everywhere.
- zope.interface.interface.Specification
  Can't be pickled. On current master, it's because the `WeakKeyDictionary` it has contains a local function which can't be pickled. With this branch, it's because objects with `tp_member` can't be pickled by default.
- zope.interface.declarations.Declaration
  Extends Specification. Can't be pickled for the same reason.

As for the base classes, they can't be pickled either (not that I think that's a use case we should support). On current master, it's because the module they name is the wrong module:
```
PicklingError: Can't pickle <class '_interface_coptimizations.SpecificationBase'>: 
    import of module '_interface_coptimizations' failed
```

(It should be `_zope_interface_coptimizations`.)

PR #155, opened 2020-01-23
More memory optimizations for a 6-7% reduction.

jamadden:
Details in commit messages. In short:

- Make Declaration, Specification, ClassProvides use `__slots__` or `tp_members`. Shouldn't affect pickling as outlined in #154. Would break sticking arbitrary objects on them, of course.
- Avoid allocating WeakKeyDictionary for dependents unless needed (< 50% of the time)
- Likewise for tagged value storage.

I think these optimizations are safe. Adding `__slots__` to `Element`, `Method` or `Attribute` leads to hard-to-resolve layout or metaclass conflicts, so I didn't touch them. Also, using `__doc__` as an instance variable makes it *really, really* hard to do anything with it.

On loading a sample application[1] of 6000 modules that defined 2200 interfaces, I got a total memory savings of about 6-7% (roughly 235MB to 223MB) from 4.7.1, as measured (repeatedly) using psutil. Startup time was essentially unchanged according to the shell's `time` command. Actual differences will depend on the application and its use of interfaces and interface features.

[1] Not an actual application, but a buildout eggs directory that *used* to belong to a Pyramid+Zope 3 application (but has also been shared with some other minor things) and so has some version of its collected dependencies (duplicates were filtered out).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,19 +11,32 @@
   value forces the Python implementation to be used, ignoring the C
   extensions. See `PR 151 <https://github.com/zopefoundation/zope.interface/pull/151>`_.
 
+- Cache the result of ``__hash__`` method in ``InterfaceClass`` as a
+  speed optimization. The method is called very often (i.e several
+  hundred thousand times during Plone 5.2 startup). Because the hash value never
+  changes it can be cached. This improves test performance from 0.614s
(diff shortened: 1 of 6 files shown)


jensens:
LGTM. That said, I am not an expert for C-extensions.

jamadden:
Thanks!

PR #159, opened 2020-01-27
Remove support for hashing uninitialized interfaces.

### LINKED ISSUE #157: Remove support for "hashing" uninitialized interfaces
Since [this commit in 2011](https://github.com/zopefoundation/zope.interface/commit/05a899a8275bbd6b2e332a4fbfeab4d7e1dc40f8) (zope.interface 3.6.5) we've been emitting a warning if a buggy subclass of `InterfaceClass` attempted to hash itself before completing its initialization. In the [original bug report](https://bugs.launchpad.net/zope.interface/+bug/811792) this was identified to be a bug in the third-party module that was doing this. Because it had a release scheduled soon and because of the vagaries of downstream packaging schedules, @tseaver added the current behaviour to workaround the issue. 

But the current behaviour is not ideal; in addition to the warning it can lead to objects not being found when their hash changes, and of course there's the warning being printed. Plus, with #156, the fallback behaviour got even uglier.

We've been emitting a warning for more than 8 years. The only test in this repo that depends on the behaviour is one that specifically invokes it. 

I hope we can drop the fallback.

jamadden:
Fixes #157

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@
  Changes
 =========
 
-4.8.0 (unreleased)
+5.0.0 (unreleased)
 ==================
 
 - Support the ``PURE_PYTHON`` environment variable at runtime instead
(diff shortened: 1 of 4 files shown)


jamadden:
Thank you!

PR #160, opened 2020-01-27
Make the singleton _empty immutable.

### LINKED ISSUE #158: `_empty` singleton is mutable
zope.interface.declarations returns a global `_empty` Declaration object in a few cases. But this object is fully mutable: its bases can be set, it can be subscribed and have dependents added. That doesn't seem safe for a global object. 

Can we either:

- Stop returning a global object from these APIs (return a new object); or 
- introduce a special subclass that's truly immutable (prohibits changing bases, adding dependents, etc).

jamadden:
Fixes #158

### DIFF
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -136,7 +136,13 @@ class _ImmutableDeclaration(Declaration):
 
     @__bases__.setter
     def __bases__(self, new_bases):
-        if new_bases:
+        # We expect the superclass constructor to set ``self.__bases__ = ()``.
+        # Rather than attempt to special case that in the constructor and allow
+        # setting __bases__ only at that time, it's easier to just allow setting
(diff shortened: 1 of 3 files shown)


src/zope/interface/declarations.py
```diff
@@ -111,6 +111,53 @@ def __add__(self, other):
...
+    def __reduce__(self):
+        return "_empty"
+
+    @property
+    def __bases__(self):
+        return ()
+
+    @__bases__.setter
+    def __bases__(self, new_bases):
+        if new_bases:
```
jensens: Why allowing setting at all? When is it the case to set an empty base here?

jamadden: The constructor of the base class sets the bases.

jensens: Ok, got it! Probably a short comment in code helps to reduce confusion of our future-selfs and others reading the code :) ?

jamadden: Sure!

Issue #55, opened 2016-09-02
zope.interface 4.3.0 broke zodbbrowser


mgedmin:
zope.interface 4.3.0 broke zodbbrowser.

Last successful build (zope.interface 4.2.0):
- https://jenkins.gedmin.as/job/zodbbrowser/658/console

First failing build (zope.interface 4.3.0 is the only version that changed):
- https://jenkins.gedmin.as/job/zodbbrowser/659/console

This is not related to #43 -- the build is still broken with zope.interface 4.3.1:
- https://jenkins.gedmin.as/job/zodbbrowser/660/console

Steps to reproduce
```
git clone https://github.com/mgedmin/zodbbrowser
cd zodbbrowser
make
bin/zodbbrowser --rw empty.fs
# now open http://localhost:8070/
```

Expected behavior:
![ekrano nuotrauka is 2016-09-02 10-02-53](https://cloud.githubusercontent.com/assets/159967/18195510/889923d6-70f4-11e6-921b-1e76fba2ed15.png)

Actual behavior:
![ekrano nuotrauka is 2016-09-02 10-04-13](https://cloud.githubusercontent.com/assets/159967/18195523/a2994fa4-70f4-11e6-8642-7b10d3d4ce69.png)

mgedmin:
I've narrowed this down to the following statement in zope.container.traversal.ContainerTraverser.browserDefault:
```python
        view_name = getSiteManager(None).adapters.lookup(
            map(providedBy, (self.context, request)), IDefaultViewName)
```

`self.context` is a zope.site.folder.Folder object.  `request` is a zope.publisher.browser.BrowserRequest instance.

zope.interface 4.2.0 returns `u'zodbbrowser'`.  zope.interface 4.3.x returns `u'index.html'`.

My [site definition](https://github.com/mgedmin/zodbbrowser/blob/ac5a00dce22ac1021a27a48bc9d2586408a32002/src/zodbbrowser/standalone.py#L56-L59) explicitly asks for
```xml
          <browser:defaultView
              for="persistent.Persistent"
              name="zodbbrowser"
              xmlns:browser="http://namespaces.zope.org/browser" />
```

jamadden:
So `browser:defaultView` winds up calling [registerAdapter](https://github.com/zopefoundation/zope.publisher/blob/master/src/zope/publisher/zcml.py#L105).

In this case that translates to

```python
[...]

PR #161, opened 2020-01-28
Remove unneeded overwrite and call to anyway inherited __hash__ method

jensens:
Remove unneeded overwrite and call to anyway inherited `__hash__` method  from `.declarations.Implements` class. 

Watching a reindex index process in ZCatalog with Py-Spy after 10k samples the TotalTime for `.adapter._lookup`  was reduced from 27.5s to 18.8s (~1.5x faster) by taking almost the same OwnTime (before 15.7s/after 13.18s). Before the removed `__hash__` had an OwnTime of 7.76s.

Overall reindex index time  shrunk from 369s to 293s (1.26x faster).

I wonder why this method was necessary? Or is it still for some edge case? 
I am sure it was not placed there w/o a reason.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,13 @@
   fields in your subclass before attempting to hash or sort it. See
   `issue 157 <https://github.com/zopefoundation/zope.interface/issues/157>`_.
 
+- Remove unneeded overwrite and call to anyway inherited `__hash__` method
+  from `.declarations.Implements` class. Watching a reindex index process in
+  ZCatalog with on a Py-Spy after 10k samples the time for `.adapter._lookup`
+  was reduced from 27.5s to 18.8s (~1.5x faster). Overall reindex index time
+  shrunk from 369s to 293s (1.26x faster). See
+  `PR 161 <https://github.com/zopefoundation/zope.interface/pull/161>`_.
+
 4.7.1 (2019-11-11)
 ==================
 
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -223,9 +223,6 @@ class Implements(Declaration):
         # This spelling works under Python3, which doesn't have cmp().
         return (n1 > n2) - (n1 < n2)
 
-    def __hash__(self):
-        return Declaration.__hash__(self)
-
     # We want equality to be based on identity. However, we can't actually
     # implement __eq__/__ne__ to do this because sometimes we get wrapped in a proxy.
     # We need to let the proxy types implement these methods so they can handle unwrapping
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -223,12 +223,12 @@ class Implements(Declaration):
         # This spelling works under Python3, which doesn't have cmp().
         return (n1 > n2) - (n1 < n2)
 
-    # We want equality to be based on identity. However, we can't actually
+    # We want equality and hashing to be based on identity. However, we can't actually
     # implement __eq__/__ne__ to do this because sometimes we get wrapped in a proxy.
     # We need to let the proxy types implement these methods so they can handle unwrapping
     # and then rely on: (1) the interpreter automatically changing `implements == proxy` into
     # `proxy == implements` (which will call proxy.__eq__ to do the unwrapping) and then
-    # (2) the default equality semantics being identity based.
+    # (2) the default equality and hashing semantics being identity based.
 
     def __lt__(self, other):
         c = self.__cmp(other)


thet:
@jamadden added it here: https://github.com/zopefoundation/zope.interface/commit/bbf1a3d4792bf5b0a3969746521b27d2a2b4bf06#diff-135050e0178a707f1f3c1fd503523f70R176

jensens:
@thet yes, I did also a git blame, but do not understand why it is needed.

jamadden:
Prior to some recent commit, there wasn't a `__hash__` that could correctly be inherited. Without defining it here, you get `TypeError`:
```python traceback
  File "//src/zope/interface/interface.py", line 669, in _wire
    from zope.interface.interfaces import IAttribute
  File "//src/zope/interface/interfaces.py", line 770, in <module>
    class ObjectEvent(object):
  File "//src/zope/interface/declarations.py", line 385, in __call__
    classImplements(ob, *self.interfaces)
  File "//src/zope/interface/declarations.py", line 323, in classImplements
    spec = implementedBy(cls)
  File "//src/zope/interface/declarations.py", line 274, in implementedByFallback
    spec = Implements.named(spec_name, *[implementedBy(c) for c in bases])
  File "//src/zope/interface/declarations.py", line 274, in <listcomp>
    spec = Implements.named(spec_name, *[implementedBy(c) for c in bases])
  File "//src/zope/interface/declarations.py", line 274, in implementedByFallback
    spec = Implements.named(spec_name, *[implementedBy(c) for c in bases])
  File "//src/zope/interface/declarations.py", line 142, in named
    inst.__init__(*interfaces)
  File "//src/zope/interface/declarations.py", line 64, in __init__
    Specification.__init__(self, _normalizeargs(interfaces))
  File "//src/zope/interface/interface.py", line 190, in __init__
    self.__bases__ = tuple(bases)
  File "//src/zope/interface/interface.py", line 214, in __setBases
    self.changed(self)
  File "//src/zope/interface/declarations.py", line 67, in changed
    Specification.changed(self, originally_changed)
  File "//src/zope/interface/interface.py", line 233, in changed
    ancestors = ro(self)
  File "//src/zope/interface/ro.py", line 64, in ro
    return _mergeOrderings([_flatten(object)])
  File "//src/zope/interface/ro.py", line 41, in _mergeOrderings
    if o not in seen:
TypeError: unhashable type: 'Implements'
```

It's worth spending a few minutes tracking down the exact change that made that unnecessary and verifying that the hashes haven't changed/are still consistent with comparisons and equality. Please hold ☎️

jamadden:
Prior to 4.3.2, `Implements` defined `__eq__`, which meant that it had to define `__hash__`, or you get the `TypeError` mentioned above. `__eq__` was defined as identity (`self is other`) and we needed the matching identity `__hash__`, `object.__hash__`. So `def __hash__(self): return Declaration.__hash__(self)` was the way that was spelled. (Today I'd see if just `__hash__ = object.__hash__` works; that should also eliminate this speed penalty.)

That was all made moot in 4.3.2 with https://github.com/zopefoundation/zope.interface/issues/55 which removed the definition of `__eq__` entirely. Without an explicit `__eq__` defined, we don't have to have an explicit `__hash__` defined to avoid the `TypeError`, we inherit object's version by default.

src/zope/interface/declarations.py
```diff
@@ -223,9 +223,6 @@ def __cmp(self, other):
         # This spelling works under Python3, which doesn't have cmp().
         return (n1 > n2) - (n1 < n2)
 
-    def __hash__(self):
-        return Declaration.__hash__(self)
-
```
jamadden: Maybe add a few words to the following comment block about also needed hashing to be identity based so that if we need to return to something like this in the future we're not confused by the `TypeError`? Maybe something like (additions in bold)

> We want equality **and hashing** to be based on identity. … (2) the default equality **and hashing** semantics being identity based.

CHANGES.rst
```diff
@@ -54,6 +54,13 @@
   fields in your subclass before attempting to hash or sort it. See
   `issue 157 <https://github.com/zopefoundation/zope.interface/issues/157>`_.
 
+- Remove unneeded overwrite and call to anyway inherited `__hash__` method
+  from `.declarations.Implements` class. Watching a reindex index process in
+  ZCatalog with on a Py-Spy after 10k samples the time for `.adapter._lookup`
+  was reduced from 27.5s to 18.8s (~1.5x faster). Overall reindex index time
+  shrunk from 369s to 293s (1.26x faster). See
```
jamadden: Is this the same reindex benchmark used for the recent `InterfaceClass.__hash__` changes in #156? I'm just curious because the numbers don't quite make sense taken together and I'm not sure how to reconcile them. Hand-waving at real-world differences, etc, is fine 😄

jensens: It is the same "benchmark" - a Py-Spy connected to a running exclusively-used and fresh started Plone instance with a copy of the production database. It is the same reindex process (but today on a slighly larger data set than last time). 

The overall  reindex Index performance increment is indeed the same as above. This is just a coincidence (I myself needed to double check too). There is also some derivation possible, this is not a standardized test. With Py-Spy it is now easy to collect data about real world run time timings. 

I can clearly state: A significant reduction is there. For most use cases with less `_lookup` calls it wont be that significant. In this case with heavy `_lookup` calls it is about 1.2x-1.3x faster. 

Thus said, Plone really overuses the ZCA at some places and even at startup or for all Dexterity operations this changes will improve the overall performance (and reduces its carbon footprint, if the servers aren't running on green power).

Issue #163, opened 2020-01-28
C/Python impl difference: C impl of IFoo.__call__ swallows *all* exceptions accessing __conform__


jamadden:
While investigating #162, I found a discrepancy in the C and Python implementations of `InterfaceBase.__call__`. In C, if accessing the `__conform__` attribute returns NULL, meaning an exception was raised, we ignore it (call `PyErr_Clear`) and continue on. In the Python version, though, we just use `getattr(obj, '__conform__', None)`. In Python 3, that only catches AttributeError.

This matters because of things like Acquisition which can raise strange exceptions. In particular, the test cases for `five.intid` have a setup that with a cycle in the acquisition path. That raises a `RuntimeError` (at least with the Python implementation) and the test fails with the `RuntimeError`, instead of continuing with the normal adapter lookup.

Ideally, the C implementation wouldn't be catching and discarding arbitrary exceptions (`MemoryError`, `KeyboardInterrupt`, etc), but that ship has probably sailed. So probably the best bet is to make the Python implementation also catch all exceptions. Sigh.

jamadden:
The same thing goes for `providedBy`: The C implementation catches all exception, the Python version sensibly catches just `AttributeError` when accessing `__providedBy__`.

And then the same thing goes for `getObjectSpecification` accessing `__provides__` and `__class__`.

PR #164, opened 2020-01-28
The _empty singleton has no-op subscribe/unsubscribe methods.

### LINKED ISSUE #162: Possible problem with _ImmutableDeclaration?
With the current master and Plone there are some failiures. One of them here:

```
Failed example:
    IKeyReference(item_c)
Expected:
    ...
    TypeError: ('Could not adapt', <SimpleItem at c>,
    <InterfaceClass zope.keyreference.interfaces.IKeyReference>)
Got:
      File "/home/jensens/.pyenv/versions/3.7.5/lib/python3.7/doctest.py", line 1329, in __run
        compileflags, 1), test.globs)
      File "<doctest README.rst[112]>", line 1, in <module>
        IKeyReference(item_c)
      File "/home/jensens/.dotfiles/.buildout/shared-eggs/zope.component-4.5-py3.7.egg/zope/component/_api.py", line 141, in adapter_hook
        return sitemanager.queryAdapter(object, interface, name, default)
      File "/home/jensens/ws/cdev/p52/src/zope.interface/src/zope/interface/registry.py", line 348, in queryAdapter
        return self.adapters.queryAdapter(object, interface, name, default)
      File "/home/jensens/ws/cdev/p52/src/zope.interface/src/zope/interface/adapter.py", line 533, in _uncached_lookup
        self._subscribe(*required)
      File "/home/jensens/ws/cdev/p52/src/zope.interface/src/zope/interface/adapter.py", line 511, in _subscribe
        r.subscribe(self)
      File "/home/jensens/ws/cdev/p52/src/zope.interface/src/zope/interface/interface.py", line 235, in subscribe
        self._dependents[dependent] = self.dependents.get(dependent, 0) + 1
    TypeError: 'NoneType' object does not support item assignment

```

After debugging I found it is probably caused by the introduction of `_ImmutableDeclaration`. 
In `Specificaton` ..
https://github.com/zopefoundation/zope.interface/blob/489328a5afc1430a064db2dba7a7a408591789f9/src/zope/interface/interface.py#L228-L247
... `subscribe` and `unsubscribe` methods are defined which modifies `_dependents`, which is set deferred on first access.
In `_ImmutableDeclaration` a subclass of `Declaration` which is a subclass of above `Specification` ...
https://github.com/zopefoundation/zope.interface/blob/489328a5afc1430a064db2dba7a7a408591789f9/src/zope/interface/declarations.py#L148-L150
... the on-access creating property is overwritten, but subscribe and unsubscribe are still in place.

Code is calling subscribe in above traceback and it fails.

I still do not understand the difference of `IKeyReference` compared to other Interfaces here, it does not look special?
https://github.com/zopefoundation/zope.keyreference/blob/master/src/zope/keyreference/interfaces.py#L31

jamadden:
Turns out they can be called in some very strange error cases. See #162 and #163 for details.

This should fix #162 (at least the provided test case, five.intid, passes now).

It also does enough work on #163 that (a) the test can be written and run in pure-python mode, which was needed to debug it and (b) five.intid runs in pure-python mode (well, with a bunch of other small hacks to Acquisition, ExtensionClass, DocumentTemplate and AccessControl), but I won't claim that it fully resolves that issue. For one thing, there are no specific tests. For another, I see more such differences.

@jensens Does this solve the issues, or at least get you further?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,12 +54,18 @@
   fields in your subclass before attempting to hash or sort it. See
   `issue 157 <https://github.com/zopefoundation/zope.interface/issues/157>`_.
 
-- Remove unneeded overwrite and call to anyway inherited `__hash__` method
-  from `.declarations.Implements` class. Watching a reindex index process in
-  ZCatalog with on a Py-Spy after 10k samples the time for `.adapter._lookup`
-  was reduced from 27.5s to 18.8s (~1.5x faster). Overall reindex index time
-  shrunk from 369s to 293s (1.26x faster). See
-  `PR 161 <https://github.com/zopefoundation/zope.interface/pull/161>`_.
+- Remove unneeded override of the ``__hash__`` method from
+  ``zope.interface.declarations.Implements``. Watching a reindex index
+  process in ZCatalog with on a Py-Spy after 10k samples the time for
+  ``.adapter._lookup`` was reduced from 27.5s to 18.8s (~1.5x faster).
+  Overall reindex index time shrunk from 369s to 293s (1.26x faster).
+  See `PR 161
+  <https://github.com/zopefoundation/zope.interface/pull/161>`_.
+
+- Make the Python implementation closer to the C implementation by
+  ignoring all exceptions, not just ``AttributeError``, during (parts
+  of) interface adaptation. See `issue 163
+  <https://github.com/zopefoundation/zope.interface/issues/163>`_.
 
 4.7.1 (2019-11-11)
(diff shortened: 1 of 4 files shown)


jensens:
The `five.intid` problem is gone. There are more to tackle. I will look at it next days. All Plone tests using this branch runs in Jenkins here. https://jenkins.plone.org/job/pull-request-5.2-3.7/819/console

jamadden:
@jensens Would you consider this meregable then?

PR #167, opened 2020-02-04
Performance optimization in _lookup _lookupAll and _subscriptions

jensens:
Micro-optimization in `.adapter._lookup`, `.adapter._lookupAll` and   `.adapter._subscriptions`: By loading components.get into a local variable  before entering the loop a bytcode "LOAD_FAST 0 (components)" in the loop can be eliminated. 

In Plone, while running all tests, average speedup of   `_lookup`s owntime is ~5x (this time measured with py-spy and 35k samples while running all tests)

### DIFF
--- a/src/zope/interface/adapter.py
+++ b/src/zope/interface/adapter.py
@@ -662,17 +662,17 @@ def _lookup(components, specs, provided, name, i, l):
     # The components.get in loops is executed 100 of 1000s times.
     # by loading get into a local variable the bytecode
     # "LOAD_FAST 0 (components)" in the loop can be eliminated.
-    cget = components.get
+    components_get = components.get
     if i < l:
         for spec in specs[i].__sro__:
-            comps = cget(spec)
+            comps = components_get(spec)
             if comps:
                 r = _lookup(comps, specs, provided, name, i+1, l)
                 if r is not None:
                     return r
     else:
         for iface in provided:
-            comps = cget(iface)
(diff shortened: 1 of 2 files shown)


src/zope/interface/adapter.py
```diff
@@ -658,16 +658,21 @@ def _convert_None_to_Interface(x):
...
 def _lookup(components, specs, provided, name, i, l):
+    # this function is called very often.
+    # The components.get in loops is executed 100 of 1000s times.
+    # by loading get into a local variable the bytecode
+    # "LOAD_FAST 0 (components)" in the loop can be eliminated.
+    cget = components.get
     if i < l:
         for spec in specs[i].__sro__:
-            comps = components.get(spec)
+            comps = cget(spec)
```
jamadden: FWIW (which is not much), when I write this type of optimization, I tend to name the local variable the same as the full method call, only using an underscore:
```python
components_get = components.get
comps = components_get(spec)
```

That serves as self-documenting code about what exactly it is we're optimizing (to me at least), and minimizes the risk of clashing with some other variable. Plus it doesn't change line lengths 😄

jensens:
> I guess Plone has some very deep+wide interface hierarchies? Did this make a runtime difference on the macro (overall) scale?

Yes it has. This really makes a difference, since lookup is called so often. In fact I already refactored code in Plone to cache lookups like `query/getAdapter` or multi-adapater and utility calls.

PR #168, opened 2020-02-04
Fixes #165 (Uninitialized self._dependents in unsubscribe)

### LINKED ISSUE #165: Uninitialized self._dependents in unsubscribe
`plone.supermodel` uses interfaces very dynamic. So, here it constructs schemas/interfaces and so it comes to an unsubscribe when the 
```
Error in test test_dynamic_schema_refreshed_on_modify_schema_policy (plone.dexterity.tests.test_fti.TestFTIEvents)
  File "/home/jensens/.pyenv/versions/3.7.5/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/home/jensens/.pyenv/versions/3.7.5/lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/home/jensens/ws/cdev/p52/src/plone.dexterity/plone/dexterity/tests/test_fti.py", line 749, in test_dynamic_schema_refreshed_on_modify_schema_policy
    DexterityFTIModificationDescription('schema_policy', '')
  File "/home/jensens/ws/cdev/p52/src/plone.dexterity/plone/dexterity/fti.py", line 561, in ftiModified
    syncSchema(model.schema, schema, overwrite=True, sync_bases=sync_bases)
  File "/home/jensens/ws/cdev/p52/src/plone.supermodel/plone/supermodel/utils.py", line 342, in syncSchema
    dest.__bases__ = tuple(bases)
  File "/home/jensens/ws/cdev/p52/src/zope.interface/src/zope/interface/interface.py", line 255, in __setBases
    b.unsubscribe(self)
  File "/home/jensens/ws/cdev/p52/src/zope.interface/src/zope/interface/interface.py", line 242, in unsubscribe
    n = self._dependents[dependent]
  File "/home/jensens/.pyenv/versions/3.7.5/lib/python3.7/weakref.py", line 396, in __getitem__
    return self.data[ref(key)]
KeyError: <weakref at 0x7fea75b7a5f0; to 'InterfaceClass' at 0x7fea75b40830 (IBlank)>
```

I propose to initialise the `_dependents` in unsubscribe first before, if this was not done before.

```
@@ -235,6 +238,8 @@ class Specification(SpecificationBase):
         self._dependents[dependent] = self.dependents.get(dependent, 0) + 1
 
     def unsubscribe(self, dependent):
+        if self._dependents is None:
+            self._dependents = weakref.WeakKeyDictionary()
         try:
             n = self._dependents[dependent]
         except TypeError:
```

I'll prepare a PR.

jensens:
a problem with multiple local interface class definitions under the same name.

see #165

### DIFF
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -370,6 +370,10 @@ class InterfaceClass(Element, InterfaceBase, Specification):
 
         self.__module__ = __module__
 
+        # local class definitions may share the same module and name,
+        # use qualname for hashes
+        self.__qualname__ = attrs.get('__qualname__', name)
+
         d = attrs.get('__doc__')
         if d is not None:
             if not isinstance(d, Attribute):
@@ -591,7 +595,7 @@ class InterfaceClass(Element, InterfaceBase, Specification):
         try:
             return self._v_cached_hash
         except AttributeError:
-            self._v_cached_hash = hash((self.__name__, self.__module__))
+            self._v_cached_hash = hash((self.__qualname__, self.__module__))
         return self._v_cached_hash
 
     def __eq__(self, other):
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -585,8 +585,8 @@ class InterfaceClass(Element, InterfaceBase, Specification):
         if other is None:
             return -1
 
-        n1 = (self.__name__, self.__module__)
-        n2 = (getattr(other, '__name__', ''), getattr(other, '__module__', ''))
+        n1 = (self.__qualname__, self.__module__)
+        n2 = (getattr(other, '__qualname__', ''), getattr(other, '__module__', ''))
 
         # This spelling works under Python3, which doesn't have cmp().
         return (n1 > n2) - (n1 < n2)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -74,6 +74,13 @@
   `_lookup`s owntime is ~5x.
   See `PR 167 <https://github.com/zopefoundation/zope.interface/pull/167>`_.
 
+- Fixes problem with local interface class definitions in functions/methods in
+  the same module using the same class name. Those where considered to be hash-wise
+  the same. Use ``__qualname__`` for the hash to fix it.
+  The ``repr`` and ``__identifier__`` are not touched, because it would break a lot
+  of tests and code which uses zope.interface and relies on the current naming.
+  See `Issue 165 <https://github.com/zopefoundation/zope.interface/issues/165>`_.
+
 
 4.7.1 (2019-11-11)
 ==================
--- a/src/zope/interface/tests/test_interface.py
+++ b/src/zope/interface/tests/test_interface.py
@@ -885,7 +885,6 @@ class InterfaceClassTests(unittest.TestCase):
         self.assertFalse(one > other)
         self.assertTrue(other > one)
 
-
 class InterfaceTests(unittest.TestCase):
 
     def test_attributes_link_to_interface(self):
@@ -1847,6 +1846,34 @@ class InterfaceTests(unittest.TestCase):
             self.assertTrue(I(c) is self)
         finally:
             adapter_hooks[:] = old_adapter_hooks
+
+    def test_local_interfaces_with_same_name_and_module_are_different(self):
+        # see https://github.com/zopefoundation/zope.interface/issues/165
+        from zope.interface import Interface
+
+        def make_IFoo_1():
+            class IFoo(Interface):
+                pass
+
+            return IFoo
+
+        def make_IFoo_2():
+            class IFoo(Interface):
+                pass
+
+            return IFoo
+
+        ifoo1 = make_IFoo_1()
+        ifoo2 = make_IFoo_2()
+        self.assertFalse(ifoo1 is ifoo2)
+        self.assertNotEqual(ifoo1, ifoo2)
+        self.assertIn(ifoo1, Interface._dependents)
+        self.assertIn(ifoo2, Interface._dependents)
+        self.asserNotEqual(
+            list(Interface._dependents).index(ifoo1),
+            list(Interface._dependents).index(ifoo2),
+        )
+        self.assertNotEqual(hash(ifoo1), hash(ifoo2))
 
 
 class AttributeTests(ElementTests):
--- a/src/zope/interface/tests/test_interface.py
+++ b/src/zope/interface/tests/test_interface.py
@@ -885,6 +885,7 @@ class InterfaceClassTests(unittest.TestCase):
         self.assertFalse(one > other)
         self.assertTrue(other > one)
 
+
 class InterfaceTests(unittest.TestCase):
 
     def test_attributes_link_to_interface(self):
@@ -1869,7 +1870,7 @@ class InterfaceTests(unittest.TestCase):
         self.assertNotEqual(ifoo1, ifoo2)
         self.assertIn(ifoo1, Interface._dependents)
         self.assertIn(ifoo2, Interface._dependents)
-        self.asserNotEqual(
+        self.assertNotEqual(
             list(Interface._dependents).index(ifoo1),
             list(Interface._dependents).index(ifoo2),
         )


src/zope/interface/interface.py
```diff
@@ -591,7 +595,7 @@ def __hash__(self):
         try:
             return self._v_cached_hash
         except AttributeError:
-            self._v_cached_hash = hash((self.__name__, self.__module__))
+            self._v_cached_hash = hash((self.__qualname__, self.__module__))
```
jamadden: I'm afraid that changing this here means that `__cmp` must also change to use `__qualname__`. As it is, right now, I think the [Python language requirement](https://docs.python.org/3/reference/datamodel.html#object.__hash__) that "objects that compare equal must have equal hash codes" is broken.

Because of that, I'm also afraid that the apparent fix could be due to misbehaving dictionaries.

src/zope/interface/interface.py
```diff
@@ -370,6 +370,10 @@ def __init__(self, name, bases=(), attrs=None, __doc__=None,
 
         self.__module__ = __module__
 
+        # local class definitions may share the same module and name,
+        # use qualname for hashes
+        self.__qualname__ = attrs.get('__qualname__', name)
```
jamadden: I'm concerned that there may be a lot of unexpected knock-on effects from defining this standard attribute now, mostly with tooling. But don't ask me exactly what they might be. 

What triggers that concern is that `__qualname__` is specifically rejected below, albeit for different reasons.

jensens:
I will add a test, planned this anyway, to show the problem. Then we have a base to discuss how to solve the problem. IMO and according to how I understand [PEP3155](https://www.python.org/dev/peps/pep-3155/) `__qualname__`  was  introduced also because of thid class of problems.

jensens:
Test added. Also in separate branch to show failure against master see https://github.com/zopefoundation/zope.interface/issues/165#issuecomment-582349967

Now this PR is green, except for Python 2.7 because it has no `__qualname__`.

src/zope/interface/interface.py
jensens: A class should define a `__qualname__` if I get PEP 3155 right. Until now Interface subclasses did not. So I do not understand why 
>  there may be a lot of unexpected knock-on effects from defining this standard attribute now
May you add more details to your concern?

jamadden: > May you add more details to your concern?

I don't have specifics, but I'm thinking of things like this:
```python
if hasattr(cls, '__qualname__'):
    # We're running on Python 3. Do Python 3 things.
else:
    # We're on Python 2 *or* we're dealing with an interface. 
    # An interface is like, but not the same as, a real class
```

Suddenly code paths that weren't ever taken with an interface before could be. What happens depends on the code.

jamadden:
Thanks for the test! It's a great start, but it doesn't seem to demonstrate issue #165 yet (it's just showing the current status quo and failing because current master doesn't match the changes in this branch).  I'll also play with it and see if I can help make a  better reproducer.

jensens:
Well, it does demonstrate why  multiple definitions of IBlank in methods of a class in the same file as locals are failing wit master in this test:
https://github.com/plone/plone.dexterity/blob/a8a234d3242f2abbf30f22cea321a3eaa299f5f8/plone/dexterity/tests/test_fti.py

Lines: 688, 714, 760 ff

jamadden:
> Well, it does demonstrate why multiple definitions of IBlank in methods of a class in the same file as locals are failing wit master in this test:

No, I don't think so. All the references *should* be weak and *should* be garbage collected. (Ok, that doesn't actually happen until a GC runs, for some reason.) Even if they're not, it shouldn't really matter. For example, this modified version of the test, which I think is closer in spirit to what the reported failure was, passes fine:
```python
    def test_local_interfaces_with_same_name_and_module_are_different(self):
        # see https://github.com/zopefoundation/zope.interface/issues/165
        from weakref import ref

        def make_IFoo_1():
            class IFoo(Interface):
                pass

            return IFoo

        def make_IFoo_2():
            class IFoo(Interface):
                pass

            return IFoo

        ifoo1 = make_IFoo_1()

        class IDest(ifoo1):
            pass

        self.assertEqual(IDest.__bases__, (ifoo1,))

        ref_ifoo1 = ref(ifoo1)
        # let ifoo1 be GC'd.
        ifoo1 = None

        ifoo2 = make_IFoo_2()
        IDest.__bases__ = (ifoo2,)
        self.assertEqual(IDest.__bases__, (ifoo2,))
        self.assertIs(IDest.__bases__[0], ifoo2)
        # But it's still here. Who is keeping it alive?
        # A gc clears that up.
        import gc
        gc.collect()

        self.assertIsNone(ref_ifoo1())
```

Hmm…exactly how old a version of zope.interface are you coming from? Sometime in the distant past, equality and hashing were identity based, not name based, maybe that's the relevant change, not anything recent.

jamadden:
Ok, starting with a git clone of `plone.dexterity` and `pip install`ing it to get `zope.interface==4.7.1` I produce the same error in `test_fti`. So this isn't master-specific, it's something older than that.

```python traceback
Error in test test_dynamic_schema_refreshed_on_modify_schema_policy (plone.dexterity.tests.test_fti.TestFTIEvents)
  File "//lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "//lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "//lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "//plone.dexterity/plone/dexterity/tests/test_fti.py", line 745, in test_dynamic_schema_refreshed_on_modify_schema_policy
    ftiModified(
  File "//plone.dexterity/plone/dexterity/fti.py", line 561, in ftiModified
    syncSchema(model.schema, schema, overwrite=True, sync_bases=sync_bases)
  File "//plone.dexterity38/lib/python3.8/site-packages/plone/supermodel/utils.py", line 340, in syncSchema
    dest.__bases__ = tuple(bases)
  File "//plone.dexterity38/lib/python3.8/site-packages/zope/interface/interface.py", line 208, in __setBases
    b.unsubscribe(self)
  File "//plone.dexterity38/lib/python3.8/site-packages/zope/interface/interface.py", line 203, in unsubscribe
    raise KeyError(dependent)
KeyError: <InterfaceClass plone.dexterity.tests.test_fti.IBlank>
```

I'll take a crack at debugging this directly.

jamadden:
Ok, if I install current master, then the tests for `plone.dexterity` run correctly. So maybe…problem solved?
```console
(plone.dexterity38) plone.dexterity [master] $ zope-testrunner --test-path . --auto-color
...omitting lots of deprecation warnings...
  Ran 143 tests with 0 failures, 0 errors, 0 skipped in 0.719 seconds.
Running plone.testing.zca.UnitTesting tests:
  Tear down zope.testrunner.layer.UnitTests in 0.000 seconds.
  Set up plone.testing.zca.UnitTesting in 0.000 seconds.
  Ran 1 tests with 0 failures, 0 errors, 0 skipped in 0.029 seconds.
Tearing down left over layers:
  Tear down plone.testing.zca.UnitTesting in 0.000 seconds.
Total: 144 tests, 0 failures, 0 errors, 0 skipped in 0.895 seconds.
```

I did have to make a tiny tweak to `plone.supermodel.utils:syncSchema`. It was assuming that if `_v_attrs` existed it was not None. That's no longer the case. `_v_attrs` is always present now but may be None.

jamadden:
> Hmm…exactly how old a version of zope.interface are you coming from?

Near as I can tell, the `plone.supermodel` tests are pulling in https://dist.plone.org/release/5.0-latest/versions.cfg which are pinned to zope.interface 3.6.7 (!).

https://dist.plone.org/release/5.1-latest/versions.cfg pins to 4.4.3, and that fails in the same way as 4.7.1.

jensens:
I always use [buildout.coredev](https://github.com/plone/buildout.coredev/) and the test script installed there. Let me see if I can fix `plone.dexterity` master standalone.

jamadden:
I tracked the failure to being introduced in 4.4.3. (Under Python 3.) The only code issue listed is https://github.com/zopefoundation/zope.interface/issues/60, which was enabling the C extensions. We knew there were behaviour differences, so that's not especially helpful. Let me check on 2.7.

jensens:
I fixed the buildout/test setup for `plone.app.dexterity`. This includes an updated TravisCI configuration. 
Now it uses zope.interface = 4.6.0 from https://raw.githubusercontent.com/zopefoundation/Zope/4.1.3/versions-prod.cfg. 
TravisCI goes green here https://travis-ci.org/plone/plone.dexterity/builds/646391854

jamadden:
On Python 2.7, again installing plone.dexterity from git master, I get no failures with zope.interface 4.7.1 *or* current master (again accounting for the `_v_attrs` issue in plone.supermodel).

jensens:
https://travis-ci.org/plone/plone.dexterity/builds/646405274 test with zope.interface master and fixed plone.supermodel - and its green.

I need to go back to buildout.coredev and find the difference. Overall, this is strange.

jensens:
I am pretty sure there is some kind of problem, but probably this is not the solution. I close this now but keep the branch. I'll come back to this later, at or after the Alpine City Strategic Sprint (Plone/Zope) we host at our office next week.

PR #169, opened 2020-02-05
Add __all__ to all modules.

### LINKED ISSUE #153: Can/Should we define __all__ for modules?
Because of the dual implementation in C and Python, and the various backwards compatibility phrasings, there's some extra cruft in many of the modules. Defining `__all__` would help clarify expected usage. See https://github.com/zopefoundation/zope.interface/pull/151#discussion_r369525607

jamadden:
Fixes #153

The items that went in each ``__all__`` are based on two thnigs.

First, what is documented:

```console
 $ rg --no-filename 'import' docs/ -trst | tr -s "[:blank:]" | sort | uniq | grep zope
 >>> from pprint import pprint
 >>> from zope.interface import *
 >>> from zope.interface import Interface, Attribute, implementer
 >>> from zope.interface import alsoProvides
 >>> from zope.interface import classImplements
 >>> from zope.interface import classImplementsOnly
 >>> from zope.interface import directlyProvidedBy
 >>> from zope.interface import directlyProvides
 >>> from zope.interface import implementedBy
 >>> from zope.interface import implementer
 >>> from zope.interface import implementer_only
 >>> from zope.interface import noLongerProvides
 >>> from zope.interface import provider
 >>> from zope.interface.adapter import AdapterRegistry
 >>> from zope.interface.declarations import Declaration
 >>> from zope.interface.declarations import InstanceDeclarations
 >>> from zope.interface.declarations import ProvidesClass
 >>> from zope.interface.declarations import named
 >>> from zope.interface.exceptions import BrokenImplementation
 >>> from zope.interface.exceptions import Invalid
 >>> from zope.interface.interface import Specification
 >>> from zope.interface.interface import adapter_hooks
 >>> from zope.interface.verify import verifyObject
 >>> import zope.interface
```

Second, some personal judgement about what the public API is that I'm more than happy to have reviewed.

I also took the opportunity to drop the unneeded `_BLANK` constant. It was used not just as a keyword arg default but also in some method bodies, so saving the global attribute lookup in favor of a constant might have some tiny performance benefits.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,13 +67,19 @@
   of) interface adaptation. See `issue 163
   <https://github.com/zopefoundation/zope.interface/issues/163>`_.
 
-- Micro-optimization in `.adapter. `, `.adapter._lookupAll` and
-  `.adapter._subscriptions`: By loading components.get into a local variable
-  before entering the loop a bytcode "LOAD_FAST 0 (components)" in the loop
-  can be eliminated. In Plone, while running all tests, average speedup of
-  `_lookup`s owntime is ~5x.
-  See `PR 167 <https://github.com/zopefoundation/zope.interface/pull/167>`_.
+- Micro-optimization in ``.adapter._lookup`` , ``.adapter._lookupAll``
+  and ``.adapter._subscriptions``: By loading ``components.get`` into
+  a local variable before entering the loop a bytcode "LOAD_FAST 0
+  (components)" in the loop can be eliminated. In Plone, while running
+  all tests, average speedup of the "owntime" of ``_lookup`` is ~5x.
+  See `PR 167
+  <https://github.com/zopefoundation/zope.interface/pull/167>`_.
 
+- Add ``__all___`` declarations to all modules. This helps tools that
+  do auto-completion and documentation and results in less cluttered
+  results. Wildcard ("*") are not recommended and may be affected. See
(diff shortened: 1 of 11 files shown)


mgedmin:
> I also took the opportunity to drop the unneeded _BLANK constant. It was used not just as a keyword arg default but also in some method bodies, so saving the global attribute lookup in favor of a constant might have some tiny performance benefits.

AFAIU the only reason it was introduced was to avoid calling the `_u('')` helper function everywhere.  It was added during Python 3 porting work before Python 3 added back support for `u''` literals.

CHANGES.rst
```diff
@@ -67,13 +67,19 @@
...
-
+- Micro-optimization in ``.adapter._lookup`` , ``.adapter._lookupAll``
+  and ``.adapter._subscriptions``: By loading ``components.get`` into
+  a local variable before entering the loop a bytcode "LOAD_FAST 0
+  (components)" in the loop can be eliminated. In Plone, while running
+  all tests, average speedup of the "owntime" of ``_lookup`` is ~5x.
+  See `PR 167
+  <https://github.com/zopefoundation/zope.interface/pull/167>`_.
+
+- Add ``__all___`` declarations to all modules. This helps tools that
```
mgedmin: Too many `_` here.

jamadden:
Thank you!

Issue #171, opened 2020-02-06
verify[Object|Class] should report all errors found


jamadden:
It's painful to work through errors one at a time when all errors could be reported at once.

PR #172, opened 2020-02-06
Fix verification for methods of builtin types with pseudo-default args on PyPy

### LINKED ISSUE #118: cannot create an IDict interface that passes verifyObject on cpython and pypy
```python
class IDict(Interface):
    def get(k, default=None):
        pass

    def pop(k, default=None):
        pass
```

It doesn't appear possible to describe the dict as an interface using z.i. The only reason it works on CPython is because it hits https://github.com/zopefoundation/zope.interface/blob/c1b436de3f36bb533634b3287a601cd81b63ccb2/src/zope/interface/verify.py#L86-L88 whereas in PyPy the methods actual `types.MethodType` types and fail to validate properly inside https://github.com/zopefoundation/zope.interface/blob/c1b436de3f36bb533634b3287a601cd81b63ccb2/src/zope/interface/verify.py#L107-L120.

In the example below for `test_default_kw` the `required` and `implemented` parameters for `pop` are:

```python
>>> required
{'positional': ('k', 'default'), 'required': ('k',), 'optional': {'default': None}, 'varargs': None, 'kwargs': None}
>>> implemented
{'positional': ('key', 'default'), 'required': ('key', 'default'), 'optional': {}, 'varargs': None, 'kwargs': None}
```

I have not yet figured out how PyPy is specifying `default` as required when it is not actually required, however they closed the bitbucket issue as "a known difference between cpython and pypy" and so I think z.i or pyramid needs a way to work around it.

Related issues:

- https://github.com/Pylons/pyramid/issues/3237
- https://bitbucket.org/pypy/pypy/issues/2685/regression-with-pyramid-zopeinterface

Example:

```python
from zope.interface import (
    implementer,
    Interface,
)
from zope.interface.verify import verifyObject

def check_iface(iface):
    @implementer(iface)
    class Dict(dict):
        pass

    d = Dict()
    verifyObject(iface, d)

def test_star_args():
    class IDict(Interface):
        def get(k, *args):
            pass

        def pop(k, *args):
            pass

    check_iface(IDict) # get implementation doesn't support variable arguments.

def test_star_args_kwargs():
    class IDict(Interface):
        def get(k, *args, **kwargs):
            pass

        def pop(k, *args, **kwargs):
            pass

    check_iface(IDict) # get implementation doesn't support keyword arguments.

def test_default_kw():  # this is also the version used in pyramid
    class IDict(Interface):
        def get(k, default=None):
            pass

        def pop(k, default=None):
            pass

    check_iface(IDict) # pop implementation requires too many arguments.

def test_default_marker():
    marker = object()
    class IDict(Interface):
        def get(k, default=marker):
            pass

        def pop(k, default=marker):
            pass

    check_iface(IDict) # pop implementation requires too many arguments.
```

jamadden:
On PyPy2, they are ignored (like on CPython), but on PyPy3 they can actually be validated.

Fixes #118

To add tests for this, some minor changes to the common interfaces that remove missing methods on Python 3 were needed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -81,6 +81,21 @@
   `issue 153
   <https://github.com/zopefoundation/zope.interface/issues/153>`_.
 
+- Fix ``verifyClass`` and ``verifyObject`` for builtin types like
+  ``dict`` that have methods taking an optional, unnamed argument with
+  no default value like ``dict.pop``. On PyPy3, the verification is
+  strict, but on PyPy2 (as on all versions of CPython) those methods
+  cannot be verified and are ignored. See `issue 118
+  <https://github.com/zopefoundation/zope.interface/issues/118>`_.
+
+- Update the common interfaces ``IEnumerableMapping``,
+  ``IExtendedReadMapping``, ``IExtendedWriteMapping``,
+  ``IReadSequence`` and ``IUniqueMemberWriteSequence`` to no longer
+  require methods that were removed from Python 3 on Python 3, such as
+  ``__setslice___``. Now, ``dict``, ``list`` and ``tuple`` properly
+  verify as ``IFullMapping``, ``ISequence`` and ``IReadSequence,``
+  respectively on all versions of Python.
+
 4.7.1 (2019-11-11)
 ==================
 
--- a/src/zope/interface/tests/test_verify.py
+++ b/src/zope/interface/tests/test_verify.py
@@ -15,12 +15,29 @@
 """
 import unittest
 
+# pylint:disable=inherit-non-class,no-method-argument,no-self-argument
 
 class Test_verifyClass(unittest.TestCase):
 
-    def _callFUT(self, iface, klass):
+    verifier = None
+
+    @classmethod
+    def setUpClass(cls):
+        # zope.testrunner doesn't call setUpClass, so if you get
+        # 'NoneType is not callable', that's why.
+        cls.verifier = staticmethod(cls._get_FUT())
+
+    @classmethod
+    def _get_FUT(cls):
         from zope.interface.verify import verifyClass
-        return verifyClass(iface, klass)
+        return verifyClass
+
+    _adjust_object_before_verify = lambda self, x: x
+
+    def _callFUT(self, iface, klass, **kwargs):
+        return self.verifier(iface,
+                             self._adjust_object_before_verify(klass),
+                             **kwargs)
 
     def test_class_doesnt_implement(self):
         from zope.interface import Interface
@@ -54,7 +71,8 @@ class Test_verifyClass(unittest.TestCase):
         from zope.interface.exceptions import BrokenImplementation
 
         class ICurrent(Interface):
-            def method(): pass
+            def method():
+                pass
 
         @implementer(ICurrent)
         class Current(object):
@@ -68,7 +86,8 @@ class Test_verifyClass(unittest.TestCase):
         from zope.interface import implementer
 
         class ICurrent(Interface):
-            def method(): pass
+            def method():
+                pass
 
         @implementer(ICurrent)
         class Current(object):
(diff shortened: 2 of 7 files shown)


jensens:
OT: While reading this changes the question comes up when we drop Python 2 support. zope.interface 6?

jamadden:
> OT: While reading this changes the question comes up when we drop Python 2 support. zope.interface 6?

I know large portions of the user base are still using Python 2. I will maintain Python 2 support for the foreseeable future.

jamadden:
Thanks!

CHANGES.rst
```diff
@@ -81,6 +81,21 @@
...
+  cannot be verified and are ignored. See `issue 118
+  <https://github.com/zopefoundation/zope.interface/issues/118>`_.
+
+- Update the common interfaces ``IEnumerableMapping``,
+  ``IExtendedReadMapping``, ``IExtendedWriteMapping``,
+  ``IReadSequence`` and ``IUniqueMemberWriteSequence`` to no longer
+  require methods that were removed from Python 3 on Python 3, such as
+  ``__setslice___``. Now, ``dict``, ``list`` and ``tuple`` properly
+  verify as ``IFullMapping``, ``ISequence`` and ``IReadSequence,``
+  respectively on all versions of Python.
```
mgedmin: +:100:

CHANGES.rst
```diff
@@ -81,6 +81,21 @@
...
+
 4.7.1 (2019-11-11)
```
mgedmin: I like two blank lines before a ReST heading, but meh.

src/zope/interface/tests/test_verify.py
```diff
@@ -15,12 +15,29 @@
...
+        # 'NoneType is not callable', that's why.
+        cls.verifier = staticmethod(cls._get_FUT())
+
+    @classmethod
+    def _get_FUT(cls):
         from zope.interface.verify import verifyClass
-        return verifyClass(iface, klass)
+        return verifyClass
+
+    _adjust_object_before_verify = lambda self, x: x
```
mgedmin: This might be more readable as a regular `def`?

src/zope/interface/tests/test_verify.py
```diff
@@ -15,12 +15,29 @@
...
+
+    def _callFUT(self, iface, klass, **kwargs):
+        return self.verifier(iface,
+                             self._adjust_object_before_verify(klass),
+                             **kwargs)
```
mgedmin: Ok, this is funky sleight of hand I don't understand, but maybe things will become clearer if I read more of the diff?

src/zope/interface/verify.py
```diff
@@ -30,21 +35,21 @@
...
     This involves:
 
-      o Making sure the candidate defines all the necessary methods
+      - Making sure the candidate defines all the necessary methods
 
-      o Making sure the methods have the correct signature
+      - Making sure the methods have the correct signature
 
-      o Making sure the candidate asserts that it implements the interface
+      - Making sure the candidate asserts that it implements the interface
```
mgedmin: Nitpick: in ReST this translates to `<blockquote><ul>`; if you want to drop the unnecessary `<blockquote`>, you need to unindent the list markers.

(I personally hate that feature of ReST, but I've learned to accept it.)

src/zope/interface/tests/test_verify.py
mgedmin: (no I did not understand why all this magic was necessary)

src/zope/interface/tests/test_verify.py
jamadden: I guess that's an idiom I've fallen into. I use it primarily in tests as a signal that it will *definitely* be overridden by subclasses, possibly at runtime. I think it began in abstract base classes as a way to avoid introducing uncovered lines.

src/zope/interface/tests/test_verify.py
jamadden: Which part seems magic? 

The idea here was that rather than redefining the entirety of `_callFUT` in the subclass (I tried a few variations on that, all of them were repetitive), the subclass just defines `adjust_object_before_verify`.

CHANGES.rst
jamadden: Possibly an artifact of repeated merges. I'll be sure it gets added.

jamadden: (In fact it is added in #173.)

src/zope/interface/verify.py
jamadden: Thanks. I'll tweak it again when I'm working on #171.

jensens:
> I know large portions of the user base are still using Python 2. I will maintain Python 2 support for the foreseeable future.

That's fine. It's not much work here.

PR #173, opened 2020-02-06
Make verification errors more readable and useful.

### LINKED ISSUE #170: Strange indentation and newlines for Invalid exceptions
While looking at #118, I noticed that the error message produced is quite strangely formatted:

```
BrokenMethodImplementation: The implementation of pop violates its contract
        because implementation requires too many arguments.
```

This comes down to the use of a multi-line string literal in `__str__`: https://github.com/zopefoundation/zope.interface/blob/0048a56bac299db7c78a9c8e52c5928e4ec06f38/src/zope/interface/exceptions.py#L64-L67

jamadden:
Eliminate the trailing newlines and blank spaces (the code called them "a stupid artifact").

Include the name of the defining interface (so the user can easily look up any requirements on the attribute) and, for methods, the expected signature (no more guessing about how many arguments are required!).

This is implemented by giving Attribute and Method useful reprs and strs. Previously, they just had the defaults.

Fixes #170

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -100,11 +100,13 @@
   and ``Method``. These contain the name of the defining interface
   and the attribute. For methods, it also includes the signature.
 
-- Change the error strings returned by ``verifyObject`` and
+- Change the error strings raised by ``verifyObject`` and
   ``verifyClass``. They now include more human-readable information
   and exclude extraneous lines and spaces. See `issue 170
   <https://github.com/zopefoundation/zope.interface/issues/170>`_.
 
+  .. caution:: This will break consumers (such as doctests) that
+               depended on the exact error messages.
 
 4.7.1 (2019-11-11)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -96,6 +96,16 @@
   verify as ``IFullMapping``, ``ISequence`` and ``IReadSequence,``
   respectively on all versions of Python.
 
+- Add human-readable ``__str___`` and ``__repr___`` to ``Attribute``
+  and ``Method``. These contain the name of the defining interface
+  and the attribute. For methods, it also includes the signature.
+
+- Change the error strings returned by ``verifyObject`` and
+  ``verifyClass``. They now include more human-readable information
+  and exclude extraneous lines and spaces. See `issue 170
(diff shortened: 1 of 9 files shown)


jensens:
Is it possible that this break existing doctests?

jamadden:
> Is it possible that this break existing doctests?

Absolutely. But doctests are inherently fragile anyway and break all the time.

jamadden:
Rebased on master after merge of #172. Also clarified previous comment after submit too soon.

docs/verify.rst
```diff
@@ -126,5 +118,93 @@ any harm:
...
+
+.. doctest::
+
+   >>> class IFoo(Interface):
+   ...    def simple(arg1): "Takes one positional argument"
+   >>> @implementer(IFoo)
+   ... class Foo(object):
+   ...    pass
+   >>> verify_foo()
+   The object <Foo...> has failed to implement interface <InterfaceClass builtins.IFoo>: The IFoo.simple(arg1) attribute was not provided.
```
mgedmin: Should it say 'attribute' or 'method' here?

jamadden: I don't think that adds much value and clutters the message.

- Methods *are* attributes; i.e., `getattr(foo, 'bar')` is the same no matter what `bar` "is"
- More specifically, `Method` is a subclass of `Attribute`.
- It's perfectly fine for an object to implement an interface method by defining it as an instance attribute: `self.interface_method = lambda a, b: …`

But I'm looking at it from an experienced implementer's point of view. Maybe it would have value for end-programmers?

CHANGES.rst
```diff
@@ -96,6 +96,16 @@
   verify as ``IFullMapping``, ``ISequence`` and ``IReadSequence,``
   respectively on all versions of Python.
 
+- Add human-readable ``__str___`` and ``__repr___`` to ``Attribute``
+  and ``Method``. These contain the name of the defining interface
+  and the attribute. For methods, it also includes the signature.
```
jensens: I would add here: "This may break tests in consuming code - in special doctests - depending on the exact output of the methods. Those tests need to be adapted."

jamadden: Will do!

PR #174, opened 2020-02-08
Make verifyObject/Class collect and raise all errors instead of only the first

### LINKED ISSUE #171: verify[Object|Class] should report all errors found

jamadden:
Fixes #171.

### DIFF
--- a/src/zope/interface/exceptions.py
+++ b/src/zope/interface/exceptions.py
@@ -20,6 +20,7 @@ __all__ = [
     'DoesNotImplement',
     'BrokenImplementation',
     'BrokenMethodImplementation',
+    'MultipleInvalid',
     # Other
     'BadImplements',
     'InvalidInterface',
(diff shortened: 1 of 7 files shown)


src/zope/interface/exceptions.py
```diff
@@ -72,19 +92,19 @@ class BrokenImplementation(Invalid, _TargetMixin):
...
         self.target = target
 
-    def __str__(self):
-        return "%s has failed to implement interface %s: The %s attribute was not provided." % (
-            self._prefix,
-            self.interface,
+
+    @property
+    def _specifics(self):
+        return " The %s attribute was not provided" % (
```
mgedmin: I see that this `_specifics` property starts with a space ...

src/zope/interface/exceptions.py
```diff
@@ -97,19 +117,49 @@ class BrokenMethodImplementation(Invalid, _TargetMixin):
...
         self.method = method
         self.mess = mess
         self.target = target
 
-    def __str__(self):
-        return "%s violates its contract in %s: %s." % (
-            self._prefix,
+    @property
+    def _specifics(self):
+        return 'violates the contract of %s because %s' % (
```
mgedmin: ... but this one doesn't.

How does one know, when overriding a `_specifics` property, whether it needs to start with a space or not?

src/zope/interface/exceptions.py
```diff
@@ -97,19 +117,49 @@ class BrokenMethodImplementation(Invalid, _TargetMixin):
...
+        self.target = target
+        self.interface = iface
+        self.exceptions = exceptions
+
+    @property
+    def _specifics(self):
+        # It would be nice to use tabs here, but that
+        # is hard to represent in doctests.
+        return '\n    ' + '\n    '.join(
+            x._specifics.strip() if isinstance(x, _TargetMixin) else(str(x))
```
mgedmin: No need to wrap `(str(x))` in parens.

src/zope/interface/exceptions.py
jamadden: Because if one gets it wrong the tests break 😄 

I agree that's annoying. I'll see about unifying it. Thank you.

src/zope/interface/exceptions.py
jamadden: I blame an overaggresive [smartparens](https://github.com/Fuco1/smartparens).

jamadden:
Both macOS Python 2.7 builds were broken this time. I don't think it's anything we did, I think it's the release of `virtualenv` 20.0.0. At least, that's the first thing I notice comparing [a failed build'](https://travis-ci.org/zopefoundation/zope.interface/jobs/648429628#L150) with [the last successful build](https://travis-ci.org/zopefoundation/zope.interface/jobs/647712899#L115). It could also be some change in the terryfy stuff.

Issue #135, opened 2019-03-13
Support typing.Protocol


liran-funaro:
[PEP-0544][1] introduce `typing.Protocol` which allows defining an interface that can be validated on runtime.
This is currently available via a non-official extension to the `typing` module called [typing-extensions][2].

Although it seems similar to zope functionality, it is very different:
1. It does not verify the arguments of the methods. Not that the implementation has the correct number of arguments and not the arguments' typing.
2. Their implementation meant for static type checking.

Can zope can utilize the already existing validation implementation to also validate Protocols on runtime.
This will allow a developer to write an interface once (via a Protocol) that could be checked statically using the existing Protocol implementation and also dynamically using zope's implementation.

  [1]: https://www.python.org/dev/peps/pep-0544/
  [2]: https://pypi.org/project/typing-extensions/

pauleveritt:
I'm not sure how I feel about protocols where the implementers don't have to say what interfaces they support. It matches duck typing, but it is hard for tools that want to go find implementations (IDEs, a registry, etc.)

Still, in some mythical future zope.interface, sitting atop that machinery would indeed give (2) above, which is a big win for those buying into mypy or PyCharm.

pauleveritt:
That said, PEP 544 (under rejected ideas) does say interfaces are conceptually a superset of protocols. So your point stands.

PR #175, opened 2020-02-10
Add more common interfaces

### LINKED ISSUE #138: Can we hook zope.interface.common up to standard library ABCs/classes?
`z.i.common.idatetime` declares appropriate `classImplements` for various stdlib classes, and the same with the exceptions in `z.i.common.interfaces`. However, `z.i.c.mapping` and `z.i.c.sequences` do not. Also, there are some things missing, such as `IBytes` and `IText` and `INumber`.

Given the existence of abstract base classes that essentially define these same interfaces, is there something we can do to ease the gap by default? In the past I've used [dolmen.builtins](https://pypi.org/project/dolmen.builtins/) for some of this.

Question is inspired by [this post of glyph](https://glyph.twistedmatrix.com/2009/02/explaining-why-interfaces-are-great.html):

> One of the irritating things about using zope.interface is that the metadata about standard objects in zope.interface.common is not hooked up to anything: IMapping.providedBy({}) returns False.  ABCs will provide that metadata in the standard library, making zope.interface that much more useful once it has been upgraded to understand the declarations that the collections and numbers modules provide.

jamadden:
Derived automatically from Python stdlib ABCs.

When done, this will fix #138. 

So far, its just the collection classes. Remaining items:

- [x] Whoops, I missed `ByteString`
- [x] The `numbers` ABCs
- [x] Other common non-ABC types, like `IText` and whatever else is missing from `dolmen.builtins`
- [x] What happens to the "legacy" interfaces in mapping.py and sequence.py? Are they left completely alone? They are probably badly out of date. Do we docs-deprecate them? Maybe they can slot-in as super-classes

### DIFF
--- a/src/zope/interface/common/__init__.py
+++ b/src/zope/interface/common/__init__.py
@@ -28,11 +28,12 @@ __all__ = [
 # pylint:disable=no-self-argument,no-method-argument
 # pylint:disable=unexpected-special-method-signature
 
-def optional(meth):
+class optional(object):
     # Apply this decorator to a method definition to make it
     # optional (remove it from the list of required names), overriding
     # the definition inherited from the ABC.
-    return _decorator_non_return
+    def __init__(self, method):
+        self.__doc__ = method.__doc__
 
 
 class ABCInterfaceClass(InterfaceClass):
@@ -144,6 +145,28 @@ class ABCInterfaceClass(InterfaceClass):
             and not self.__is_reverse_protocol_name(k)
         }
 
+        methods['__doc__'] = self.__create_class_doc(attrs)
+        # Anything specified in the body takes precedence.
+        methods.update(attrs)
+        InterfaceClass.__init__(self, name, bases, methods)
+        self.__register_classes()
+
+    @staticmethod
+    def __optional_methods_to_docs(attrs):
+        optionals = {k: v for k, v in attrs.items() if isinstance(v, optional)}
+        for k in optionals:
+            attrs[k] = _decorator_non_return
+
+        if not optionals:
+            return ''
+
+        docs = "\n\nThe following methods are optional:\n - " + "\n-".join(
+            "%s\n%s" % (k, v.__doc__) for k, v in optionals.items()
+        )
+        return docs
+
+    def __create_class_doc(self, attrs):
+        based_on = self.__abc
         def ref(c):
             mod = c.__module__
             name = c.__name__
@@ -159,20 +182,18 @@ class ABCInterfaceClass(InterfaceClass):
         if implementations_doc:
             implementations_doc = "\n\nKnown implementations are:\n\n - " + implementations_doc
 
-        methods['__doc__'] = """Interface for the ABC `%s.%s`.%s""" % (
-            based_on.__module__,
-            based_on.__name__,
+        based_on_doc = (based_on.__doc__ or '')
+        based_on_doc = based_on_doc.splitlines()
+        based_on_doc = based_on_doc[0] if based_on_doc else ''
+
+        doc = """Interface for the ABC `%s.%s`.\n\n%s%s%s""" % (
+            based_on.__module__, based_on.__name__,
+            attrs.get('__doc__', based_on_doc),
+            self.__optional_methods_to_docs(attrs),
             implementations_doc
         )
-        # Anything specified in the body takes precedence.
-        # This lets us remove things that are rarely, if ever,
-        # actually implemented. For example, ``tuple`` is registered
-        # as an Sequence, but doesn't implement the required ``__reversed__``
-        # method, but that's OK, it still works with the ``reversed()`` builtin
-        # because it has ``__len__`` and ``__getitem__``.
-        methods.update(attrs)
-        InterfaceClass.__init__(self, name, bases, methods)
-        self.__register_classes()
+        return doc
+
 
     @staticmethod
     def __is_private_name(name):
--- a/src/zope/interface/common/__init__.py
+++ b/src/zope/interface/common/__init__.py
@@ -210,20 +210,25 @@ class ABCInterfaceClass(InterfaceClass):
 
     def getRegisteredConformers(self):
         """
-        Return an iterable of the classes that are directly
-        registered to conform to the ABC this interface
-        parallels.
+        Return an iterable of the classes that are known to conform to
+        the ABC this interface parallels.
         """
         based_on = self.__abc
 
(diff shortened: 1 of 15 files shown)


jamadden:
This still leaves classes in control of what interfaces they implement, unlike the virtual duck-typed subclassing of ABCs, as described in #135 

It gets us most of the way there, though. If we want to do it, it's now pretty easy with a small change to the definition of `implementedBy` (and maybe a corresponding tweak to `ABCInterface.providedBy` for direct queries; not shown here):

```patch
-------------------------------------------------------------------------------------------------
modified: src/zope/interface/declarations.py
-------------------------------------------------------------------------------------------------
@@ -275,6 +275,27 @@ def _implements_name(ob):
     return (getattr(ob, '__module__', '?') or '?') + \
         '.' + (getattr(ob, '__name__', '?') or '?')

+def _abc_interfaces():
+    try:
+        from zope.interface.common import ABCInterface, ABCInterfaceClass
+    except ImportError:
+        return [] # Circular imports at _wire() time.
+    from zope.interface.common import collections, numbers, io # Ensure defined
+
+    seen = set()
+    to_examine = list(ABCInterface.dependents)
+    while to_examine:
+        iface = to_examine.pop()
+        if iface in seen:
+            continue
+        seen.add(iface)
+        to_examine.extend(iface.dependents)
+    return [x for x in seen if isinstance(x, ABCInterfaceClass)]
+
+def _abc_interfaces_of(cls):
+    ifaces = _abc_interfaces()
+
+    return [iface for iface in ifaces if issubclass(cls, iface.getABC())]

 @_use_c_impl
 def implementedBy(cls):
@@ -338,7 +359,8 @@ def implementedBy(cls):
                 raise TypeError("ImplementedBy called for non-factory", cls)
             bases = ()

-        spec = Implements.named(spec_name, *[implementedBy(c) for c in bases])
+        spec = Implements.named(spec_name,
+                                *([implementedBy(c) for c in bases] + _abc_interfaces_of(cls)))
         spec.inherit = cls

     try:
```

With that change, classes that "subclass" ABCs automatically also provide those interfaces:

```pycon
>>> from zope.interface.common import collections
>>> class Sized(object):
   ...:     def __len__(self): return 0
   ...:
>>> list(providedBy(Sized()))
[<ABCInterfaceClass zope.interface.common.collections.IHashable>,
 <ABCInterfaceClass zope.interface.common.collections.ISized>]
```

That's a rough approximation, of course. The biggest issue is later registrations and cache invalidations. E.g., this doesn't do what you'd expect:
```pycon
>>> class VirtualSized(object):
...     pass
...
>>> list(providedBy(VirtualSized()))
[<ABCInterfaceClass zope.interface.common.collections.IHashable>]
>>> from collections.abc import Sized
>>> Sized.register(VirtualSized)
<class '__main__.VirtualSized'>
>>> isinstance(VirtualSized(), Sized)
True
>>> list(providedBy(VirtualSized()))
[<ABCInterfaceClass zope.interface.common.collections.IHashable>]
```

I feel like that's solvable, though. 

Is that something we want to proceed with? I'm not sure I personally have a defined use-case for it right now, but OTOH, I think I remember writing things like `classImplements(OOBTree, IMapping); classImplements(dict, IMapping)` just to be able to use the same adapters for both `OOBTree` and `dict`, and this would make that much more convenient (except for some reason, `OOBTree` isn't actually a virtual subclass of `collections.abc.Mapping` 😞 ).

jensens:
Overall this looks fine. I must admit, I can not judge every detail.

src/zope/interface/common/__init__.py
```diff
@@ -1,2 +1,259 @@
...
+        None
+        >>> list(providedBy(SizedClass()))
+        []
+
+    Thus, to avoid conflicting assumptions, ABCs should not be
+    declared to implement their parallel ABC interface. Only concrete
+    classes specifically registered with the ABC should be declared to
+    do so.
+
+    .. verisonadded:: 5.0
```
ale-rt: `verisonadded` -> `versionadded`

src/zope/interface/common/numbers.py
```diff
@@ -0,0 +1,83 @@
...
+
+class IComplex(INumber):
+    abc = abc.Complex
+
+    @optional
+    def __complex__():
+        """
+        Rarely implemented, even in builtin types.
+        """
+    if PY2:
```
ale-rt: Isn't a newline missing above this `if PY2`?

src/zope/interface/common/__init__.py
ale-rt: Nitpicking: here the PR mentions 5.0, later 5.0.0

src/zope/interface/common/io.py
```diff
@@ -0,0 +1,52 @@
...
+
+import io as abc
+try:
+    import cStringIO
+    import StringIO
+except ImportError:
+    # Python 3
+    extra_buffered_io_base = ()
+else:
+    extra_buffered_io_base = (StringIO.StringIO, cStringIO.InputType, cStringIO.OutputType)
```
ale-rt: I would not add the `extra_buffered_io_base` module variable if possible, I'd keep its definition in the `IBufferedIOBase` class definition. Not a blocker.

src/zope/interface/common/tests/test_builtins.py
```diff
@@ -0,0 +1,59 @@
...
+from zope.interface._compat import PYTHON2 as PY2
+from zope.interface.common import builtins
+
+from . import VerifyClassMixin
+from . import VerifyObjectMixin
+
+
+class TestVerifyClass(VerifyClassMixin,
+                      unittest.TestCase):
+    UNVERIFIABLE = (
```
ale-rt: `UNVERIFIABLE = ()` looks more consistent (it is used later in this patch) and friendly to code formatters.
Also I see later it is used as a dictionary.
So maybe it should be defined as `UNVERIFIABLE = {}`.

src/zope/interface/common/io.py
jamadden: Good idea!

jamadden:
Thanks for the feedback!

mgedmin:
This is rather a big diff and I'm not feeling up to reviewing it right now.

(Lack of review from little old me should not be a blocker to merging!)

jamadden:
> This is rather a big diff and I'm not feeling up to reviewing it right now.

No worries! Thank you.

I will go ahead and merge as I'd like to make a release soon. Since it's completely opt-in I don't think it can do any harm.

PR #176, opened 2020-02-14
Add test case for #6.

### LINKED ISSUE #6: Assignment to __class__ is broken
In https://bugs.launchpad.net/zope.interface/+bug/180844, @jimfulton reports:

> See: http://mail.zope.org/pipermail/interface-dev/2008-January/000192.html
> 
> Here's an example that provides the basis for a test case:

```pycon
>>> import zope.interface.interface
>>> class InterfaceClass(zope.interface.interface.InterfaceClass):
...     def __call__(self, *args):
...         print 'called', self, args
...
>>> IFoo = InterfaceClass('IFoo', (zope.interface.Interface,))
>>> IFoo(1)
called <InterfaceClass __main__.IFoo> (1,)
>>> class IBar(IFoo): pass
...
>>> IBar()
called <InterfaceClass __main__.IBar> ()
>>> class ISpam(zope.interface.Interface): pass
...
>>> ISpam()
    File "<console>", line 1, in ?
TypeError: function takes at least 1 argument (0 given)
>>> ISpam.__class__ = InterfaceClass
>>> ISpam()
    File "<console>", line 1, in ?
TypeError: function takes at least 1 argument (0 given)
```

jamadden:
It seems to be working now, so this fixes #6

I tried to reproduce the example by hand in Python 3.8 and 2.7 and couldn't. I don't know what I saw when I commented earlier that it was failing with the C accelerators.

### DIFF
--- a/src/zope/interface/tests/test_interface.py
+++ b/src/zope/interface/tests/test_interface.py
@@ -418,7 +418,7 @@ class InterfaceClassTests(unittest.TestCase):
         from zope.interface.interface import InterfaceClass
         return InterfaceClass
 
-    def _makeOne(self,  name='ITest', bases=(), attrs=None, __doc__=None,
+    def _makeOne(self, name='ITest', bases=(), attrs=None, __doc__=None,
                  __module__=None):
         return self._getTargetClass()(name, bases, attrs, __doc__, __module__)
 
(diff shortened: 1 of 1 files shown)


jamadden:
Yeah, I was thinking since it's tests only it didn't really need a note.

PR #177, opened 2020-02-15
Fix repr of "Attribute" to look like before

jensens:
and as usual in Python.

It broke tons of tests in Plone and related. Also the repr looks like it's usual in Python.


jensens:
indeed, I'll remove the note at all, as it is a fix for the newly introduced change.

PR #179, opened 2020-02-16
performance: store cached hashvalue in slot

jensens:
Micro-optimizations. In fact this does only save a few percent: 
- Py-Spy on Plone, running all tests.
- 20k samples (200sec)
- with current master `__hash__` takes 2.6% of time
- with this branch `__hash__` takes 1.6% of time

I did not found much information on slots, Python C-extensions and inheritance. I hope I got it right.


jensens:
Scrap it, if something helps , then a better algorythm/C-impl.

PR #180, opened 2020-02-18
Implement _lookup in C-extension

jensens:
At the moment this is tested only with Python 3.7 and may fail on other versions.

Speed-up is only ~1.7x, I would have expected more. 
A 200s sample with Py-Spy running the all Plone tests comes down from 11.1s for _lookup to 6.5s.

Todo: 
- [ ] Make it work with all here supported Python versions.
- [ ] Better performance checking code than using Py-Spy
- [ ] Optimize code if possible
- [ ] Review reference-counting.

Feedback appreciated, my C is a bit rusty (not much used the last 20 years) and I never wrote C-extension code before.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -116,6 +116,10 @@
   Like the above, this will break consumers depending on the exact
   output of error messages if more than one error is present.
 
+- Implement ``adapter._lookup`` in C. In a Plone test run measured with
+  Py-Spy the function is 1.7x faster than before. See `PR xxx
+  <https://github.com/zopefoundation/zope.interface/pull/xxx>`_.
+
 
 4.7.1 (2019-11-11)
 ==================
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -1678,6 +1678,125 @@ static PyTypeObject VerifyingBase = {
 /* ==================================================================== */
 
 
+/* ==================================================================== */
+/* ========================== Begin: _recursive_lookup ================ */
+
+// def _lookup(components, specs, provided, name, i, l):
+//     components_get = components.get
+//     if i < l:
+//         for spec in specs[i].__sro__:
+//             comps = components_get(spec)
+//             if comps:
+//                 r = _lookup(comps, specs, provided, name, i+1, l)
+//                 if r is not None:
+//                     return r
+//     else:
+//         for iface in provided:
+//             comps = components_get(iface)
+//             if comps:
+//                 r = comps.get(name)
+//                 if r is not None:
+//                     return r
+//     return None
+
+
+static PyObject*
+_lookup_recursive(PyObject *components,
+                  PyObject *specs,
+                  PyObject *provided,
+                  PyObject *name,
+                  int i,
+                  int l)
+{
+  /* recursive component lookup
+  */
+  PyObject *result=NULL,
+           *comps=NULL,
+           *sro_at_i=NULL,
+           *key=NULL;
+  Py_hash_t hash;
+  PyListObject *provided_list;
+  int idx, length;
+
+  if (i < l) {
+    sro_at_i = PyObject_GetAttrString(PyTuple_GetItem(specs, i), "__sro__");
+    length = PySequence_Size(sro_at_i);
+    for (idx = 0; idx < length; idx++) {
+      key = ((PyTupleObject *)sro_at_i) -> ob_item[idx];
+      hash = PyObject_Hash(key);
+      if (hash != -1) {
+        comps = _PyDict_GetItem_KnownHash(components, key, hash);
+        if (comps && comps!=Py_None) {
+          result = _lookup_recursive(comps, specs, provided, name, i+1, l);
+          if (result && result!=Py_None) {
+              return result;
+          }
+        }
+      }
+    }
+  } else {
+    provided_list = (PyListObject *)provided;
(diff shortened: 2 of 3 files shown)


jensens:
see also here https://community.plone.org/t/profiling-tests-in-plone-found-zope-interface-as-bottleneck/11714

jamadden:
I wonder how good (realistic) a benchmark a large test suite is, though. A lot of how zope.interface expects to gain performance is through extensive caching, and test suites are known to be *bad* for caching. They register and unregister adapters, they create and throw away new interfaces, etc. (These are the same sorts of reasons that PyPy's JIT [may not do well on test suites.](https://www.pypy.org/performance.html)) 

Observe that `_lookup` is called from `_uncached_lookup`, which is not the expected fast path. Yes, using a test suite can help find and optimize certain hotspots in the slow paths of the code, and that will have some performance benefit overall, but unless the test suite is exercising the expected fast paths that a real application does, the benefits may be limited to just tests.

jamadden:
If you're running on a Mac, or don't specify the `—native` option to py-spy, py-spy can't "see" native functions. So the time spent running them gets added to the time spent in the caller. If the bulk of the time spent in a function was in a call to *another* function, then you'll see a huge jump in the `OwnTime` of that function when you move the called function into C; `TotalTime` is the more direct measurement. We seem to see that here. 

With `_lookup` in Python:

```
%Own   %Total  OwnTime  TotalTime  Function (filename)
0.00%   0.00%   96.23s    125.7s   _lookup (zope/interface/adapter.py)
0.00%   0.00%   36.35s    36.35s   __hash__ (zope/interface/interface.py)
0.00%   0.00%    4.64s    132.3s   _uncached_lookup (zope/interface/adapter.py)

real	21m20,264s
user	20m54,982s
sys	0m43,547s
```

The 132s of `_uncached_lookup` (the only caller of `_lookup`)  is about 10% of 21:20. Or, the 96s of `_lookup` is about 7% of the duration.

With `_lookup` in C:
```
%Own   %Total  OwnTime  TotalTime  Function (filename)
0.00%   0.00%   63.13s    96.56s   _uncached_lookup (zope/interface/adapter.py)
0.00%   0.00%   39.80s    39.80s   __hash__ (zope/interface/interface.py)


real	23m46,821s
user	23m17,787s
sys	0m52,630s
```

The 97s attributed to `_uncached_lookup` is about 6.8% of 23:46. (We have to continue to use the `TotalTime` to get a direct comparison.) The `OwnTime` is about 4.5%. 

So the overall improvement is no more than 3%. That doesn't look much like a bottleneck to me, it's more the "death by a thousand cuts" type scenario 😄 

There may be some benefits to be had by carefully using Cython to compile most of the modules. It takes work and careful type declarations, but it's still easier, safer, and more portable than pure C.

jensens:
> Observe that _lookup is called from _uncached_lookup, which is not the expected fast path. Yes, using a test suite can help find and optimize certain hotspots in the slow paths of the code, and that will have some performance benefit overall, but unless the test suite is exercising the expected fast paths that a real application does, the benefits may be limited to just tests.

I observed the very same bottlenecks in a live site. My "very real world" test" there was a reindex catalog in Plone with ~50k objects. So while I agree in general that a test suite is not the perfect setup for performance testing: It matches good enough with my real world in-production Plone site observations. 

I am open for any better/easier solution. Cython also sounds like worth a try - another skill I have to  to learn. I must admit, I do not like writing C-extensions using Python C-API with it's confusing manual reference counting. It is easy to introduce bugs and memory leaks. 

Another question here is, if it is possible to optimize the lookup algorithm. The expensive parts here are the very frequent dictionary lookups. Our custom hashing method add up here too.

jensens:
> If you're running on a Mac, or don't specify the —native option to py-spy, py-spy can't "see" native functions.

Yes, but native set Py-Spy did not work properly here on my Linux. The differences between two of the same runs were large. On a by-line output we can observe the shift from `_lookup` function to the line in `_uncached_lookup` where the native `_lookup` is called with its 63.13s.

jensens:
> The 97s attributed to _uncached_lookup is about 6.8% of 23:46. (We have to continue to use the TotalTime to get a direct comparison.) The OwnTime is about 4.5%.

Sorry, I dont get your calculations if using TotalTime as you proposed. The overall performance (TotalTime) of `_uncached_lookup` including  `_lookup` and `__hash__` gets down from 132s to 96s - which is 73% of the pure Python implementation or about 1.36x faster.

jamadden:
> down from 132s to 96s - which is 73% of the pure Python implementation or about 1.36x faster.

In isolation, yes. But two things reduce the impact when you look at on a holistic basis. First, the overall run took 2:26 longer(12%) and that has to be taken into account. 

Second, just because one thing in isolation is faster doesn't mean that it had that much impact on the overall process; it's kind of Amdahl's law, except non-parallel (there must be a name but I'm blanking at the moment). The fastest code is the code that doesn't run, so even if we take that 96s all the way down to zero, the overall performance improvement isn't 100%, it's closer to 9%, not 1.36x, because the original code didn't take 36% of the total time.

jamadden:
> if it is possible to optimize the lookup algorithm. The expensive parts here are the very frequent dictionary lookups.

It's pretty hard to beat `O(1)` complexity 😄 There are some `O(n)` parts that perhaps could be improved, but it's non-obvious to me how.

jensens:
I close this: a) @jamadden is right b) no resources to dig deeper

PR #181, opened 2020-03-05
Make provided/implementedBy and adapter registries respect super().

### LINKED ISSUE #11: `providedBy` does not respect `super`
In https://bugs.launchpad.net/zope.interface/+bug/448712, Dieter Maurer reported:

> In some places, the implementation of a subclass adapter may take advantage of that defined for a base class. In such cases, "super" comes to mind. Unfortunately, "providedBy" (implicitely during in adapter lookup) does not respect "super" as shown by the folloging transscript:

```
>>> from zope.interface import Interface, implements
>>> class I1(Interface): pass
...
>>> class B(object): implements(Interface)
...
>>> class C(B): implements(I1)
...
>>> c=C()
>>> list(providedBy(c))
[<InterfaceClass __main__.I1>, <InterfaceClass zope.interface.Interface>]
>>> list(providedBy(super(C,c)))
[<InterfaceClass __main__.I1>, <InterfaceClass zope.interface.Interface>]
```

jamadden:
The query functions now start by looking at the next class in the MRO (interfaces directly provided by the underlying object are not found).

Adapter registries automatically pick up providedBy change to start finding the correct implementations of adapters, but to make that really useful they needed to change to unpack super() arguments and pass `__self__` to the factory.

Fixes #11

Unfortunately, this makes PyPy unable to build the C extensions.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -134,6 +134,23 @@
   `dolmen.builtins <https://pypi.org/project/dolmen.builtins/>`_.
   See `issue 138 <https://github.com/zopefoundation/zope.interface/issues/138>`_.
 
+- Make ``providedBy()`` and ``implementedBy()`` respect ``super``
+  objects. For instance, if class ``Derived`` implements ``IDerived``
+  and extends ``Base`` which in turn implements ``IBase``, then
+  ``providedBy(super(Derived, derived))`` will return ``[IBase]``.
+  Previously it would have returned ``[IDerived]`` (in general, it
+  would previously have returned whatever would have been returned
+  without ``super``).
+
+  Along with this change, adapter registries will unpack ``super``
+  objects into their ``__self___`` before passing it to the factory.
+  Together, this means that ``component.getAdapter(super(Derived,
+  self), ITarget)`` is now meaningful.
+
+  See `issue 11 <https://github.com/zopefoundation/zope.interface/issues/11>`_.
+
+- Fix a potential interpreter crash in the low-level adapter
+  registry lookup functions. See issue 11.
 
 4.7.1 (2019-11-11)
 ==================
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -38,16 +38,12 @@
 static PyObject *str__dict__, *str__implemented__, *strextends;
 static PyObject *BuiltinImplementationSpecifications, *str__provides__;
 static PyObject *str__class__, *str__providedBy__;
-static PyObject *empty, *fallback, *str_implements;
+static PyObject *empty, *fallback;
 static PyObject *str__conform__, *str_call_conform, *adapter_hooks;
 static PyObject *str_uncached_lookup, *str_uncached_lookupAll;
 static PyObject *str_uncached_subscriptions;
 static PyObject *str_registry, *strro, *str_generation, *strchanged;
-static PyObject *str__get__;
 static PyObject *str__self__;
-static PyObject *str__thisclass__;
-static PyObject *str__self_class__;
-static PyObject *str__mro__;
 
 static PyTypeObject *Implements;
 
@@ -1045,7 +1041,7 @@ lookup_lookup(lookup *self, PyObject *args, PyObject *kwds)
   static char *kwlist[] = {"required", "provided", "name", "default", NULL};
   PyObject *required, *provided, *name=NULL, *default_=NULL;
 
-  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO", kwlist,
+  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO:LookupBase.lookup", kwlist,
                                     &required, &provided, &name, &default_))
     return NULL;
 
@@ -1117,7 +1113,7 @@ lookup_lookup1(lookup *self, PyObject *args, PyObject *kwds)
   static char *kwlist[] = {"required", "provided", "name", "default", NULL};
   PyObject *required, *provided, *name=NULL, *default_=NULL;
 
-  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO", kwlist,
+  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO:LookupBase.lookup1", kwlist,
                                     &required, &provided, &name, &default_))
     return NULL;
 
@@ -1133,6 +1129,8 @@ lookup_lookup1(lookup *self, PyObject *args, PyObject *kwds)
             factory = self.lookup((required, ), provided, name)
 
         if factory is not None:
+            if isinstance(object, super):
+                object = object.__self__
             result = factory(object)
             if result is not None:
                 return result
@@ -1201,7 +1199,7 @@ lookup_adapter_hook(lookup *self, PyObject *args, PyObject *kwds)
   static char *kwlist[] = {"provided", "object", "name", "default", NULL};
   PyObject *object, *provided, *name=NULL, *default_=NULL;
 
-  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO", kwlist,
+  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO:LookupBase.adapter_hook", kwlist,
                                     &provided, &object, &name, &default_))
     return NULL;
 
@@ -1214,7 +1212,7 @@ lookup_queryAdapter(lookup *self, PyObject *args, PyObject *kwds)
   static char *kwlist[] = {"object", "provided", "name", "default", NULL};
   PyObject *object, *provided, *name=NULL, *default_=NULL;
 
-  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO", kwlist,
+  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO|OO:LookupBase.queryAdapter", kwlist,
                                     &object, &provided, &name, &default_))
     return NULL;
 
@@ -1285,7 +1283,7 @@ lookup_lookupAll(lookup *self, PyObject *args, PyObject *kwds)
   static char *kwlist[] = {"required", "provided", NULL};
   PyObject *required, *provided;
 
-  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO", kwlist,
+  if (! PyArg_ParseTupleAndKeywords(args, kwds, "OO:LookupBase.lookupAll", kwlist,
                                     &required, &provided))
     return NULL;
 
@@ -1752,7 +1750,6 @@ init(void)
   DEFINE_STRING(__class__);
   DEFINE_STRING(__providedBy__);
   DEFINE_STRING(extends);
-  DEFINE_STRING(_implements);
   DEFINE_STRING(__conform__);
   DEFINE_STRING(_call_conform);
   DEFINE_STRING(_uncached_lookup);
@@ -1763,10 +1760,6 @@ init(void)
   DEFINE_STRING(ro);
   DEFINE_STRING(changed);
   DEFINE_STRING(__self__);
-  DEFINE_STRING(__get__);
-  DEFINE_STRING(__thisclass__);
-  DEFINE_STRING(__self_class__);
-  DEFINE_STRING(__mro__);
 #undef DEFINE_STRING
   adapter_hooks = PyList_New(0);
   if (adapter_hooks == NULL)
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -43,6 +43,11 @@ static PyObject *str__conform__, *str_call_conform, *adapter_hooks;
 static PyObject *str_uncached_lookup, *str_uncached_lookupAll;
 static PyObject *str_uncached_subscriptions;
 static PyObject *str_registry, *strro, *str_generation, *strchanged;
+static PyObject *str__get__;
+static PyObject *str__self__;
+static PyObject *str__thisclass__;
+static PyObject *str__self_class__;
+static PyObject *str__mro__;
 
 static PyTypeObject *Implements;
 
@@ -91,6 +96,7 @@ import_declarations(void)
   return 0;
 }
 
+
 static PyTypeObject SpecType;   /* Forward */
 
 static PyObject *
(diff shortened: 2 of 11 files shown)


icemac:
@d-maurer I added you as reviewer as #11 was originally written by you.

CHANGES.rst
```diff
@@ -134,6 +134,20 @@
...
+  and extends ``Base`` which in turn implements ``IBase``, then
+  ``providedBy(super(Derived, derived))`` will return ``[IBase]``.
+  Previously it would have returned ``[IDerived]`` (in general, it
+  would previously have returned whatever would have been returned
+  without ``super``).
+
+  Along with this change, adapter registries will unpack ``super``
+  objects into their ``__self___`` before passing it to the factory.
+  Together, this means that ``component.getAdapter(super(Derived,
+  self), ITarget)`` is now meaningful.
```
d-maurer: While the unpacking is sufficient for my original use case, this may not always be the case, i.e. in some cases, one might want that the super object (and not its `__self__`) is passed to the adapter. I agree that given the current interface, it is impossible to express what the use case wants. In the long term, I would like the interface to be enhanced by a `generateAdapterFactories` decoupling adapter factory lookup and calling them. This would also significantly facilitate the implementation of "conditional adapters" (a concept, I have used for `Products.AdvancedQeury`).

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -91,6 +95,41 @@ import_declarations(void)
...
+    result = class_that_invoked_super = mro = NULL;
+    #define _NSCHECK(ob, bad_val) do { if (ob == bad_val) { goto EXIT; } } while (0)
+
+    class_that_invoked_super = PyObject_GetAttr(ob, str__thisclass__);
+    _NSCHECK(class_that_invoked_super, NULL);
+
+    mro = PyObject_CallMethodObjArgs(class_that_invoked_super, strmro, NULL);
+    _NSCHECK(mro, NULL);
+
+    index = PySequence_Index(mro, class_that_invoked_super);
```
d-maurer: I think that the `mro` must be determined from `ob.__self_class__` not from `ob.__thisclass__`. I expect that `index` above will always be `0`.

src/zope/interface/declarations.py
```diff
@@ -54,6 +54,15 @@ class implements (that instances of the class provides).
 _ADVICE_WARNING = ('The %s API is deprecated, and will not work in Python3  '
                    'Use the @%s class decorator instead.')
 
+def _next_super_class(ob):
+    # When ``ob`` is an instance of ``super``, return
+    # the next class in the MRO that we should actually be
+    # looking at.
+    class_that_invoked_super = ob.__thisclass__
+    mro = class_that_invoked_super.mro()
+    next_class = mro[mro.index(class_that_invoked_super) + 1]
```
d-maurer: Again: the `mro` should be determined for `ob.__self_class_`, not `ob.__thisclass__`.

src/zope/interface/_zope_interface_coptimizations.c
jamadden: I'm afraid you have that backwards. In `super(Class, self)` `__self_class__` is always the same as `type(self)`. It never changes and is the most derived class in the MRO (index 0).  It's `__thisclass__` that changes to reflect where in the MRO the new starting search point actually is.

```pycon
>>> class A(object):
...     pass
...
>>> class B(A):
...     pass
...
>>> class C(B):
...     pass
...
>>> class D(C):
...     pass
...
>>> D.__mro__
(<class '__main__.D'>,
 <class '__main__.C'>,
 <class '__main__.B'>,
 <class '__main__.A'>,
 <class 'object'>)
>>> d = D()
>>> sd = super(D, d)
>>> sc = super(C, d)
>>> sb = super(B, d)

>>> (D, sd.__self_class__, sc.__self_class__, sb.__self_class__)
(<class '__main__.D'>,
 <class '__main__.D'>,
 <class '__main__.D'>,
 <class '__main__.D'>)

>>> sd.__thisclass__
<class '__main__.D'>
>>> sc.__thisclass__
<class '__main__.C'>
>>> sb.__thisclass__
<class '__main__.B'>
```

CHANGES.rst
jamadden: I don't have a strong immediate use case, but it seemed to me that passing the `super` object was both not very useful and inconsistent with other uses of both `super`. The practical uses I could think of for today all work best with `__self__` unwrapped.

It's inconsistent because when `super(Class, self).method()` is executing, the `self` that's sent to `method` isn't the `super` object, it's straight up exactly the `self` object provided to `super` (because `super` isn't particularly useful; see next paragraph).  Adapting on `super` ought to implicitly behave the same simply to match that expectation.

It's not particularly useful because the `super` object cannot access instance variables of `self` (only class descriptors/methods), nor (as a consequence) can it discover any interfaces that are `directlyProvidedBy` the `self` object. It has no MRO or type and `super` doesn't work on it, so you'd only be able to try this trick for exactly one level of hierarchy. The receiving factory would need to unwrap `__self__` in short order to do any of those things…and at that point, you're writing an adapter that's customized to know it's working with `super`, which destroys reusability: given `class Base` and `class Derived(Base)` the same adapter can't work for `super(Derived, self)` *and* `Base()`.

If one wanted to use the `super` object in an adapter, I think the better thing to do would be to register a multi-adapter on `(Interface-or-class, type)` (called with `component.getMultiAdapter(IAdapter, (self, Derived))`) or possibly `(Interface-or-class, super)`. The latter would work today (but not after this change), while the former would work today and after this change.

d-maurer:
Jason Madden wrote at 2020-3-7 01:44 -0800:
>jamadden commented on this pull request.
>
>
>
>> +    mro = PyObject_CallMethodObjArgs(class_that_invoked_super, strmro, NULL);
>+    _NSCHECK(mro, NULL);
>+
>+    index = PySequence_Index(mro, class_that_invoked_super);
>
>I'm afraid you have that backwards. In `super(Class, self)` `__self_class__` is always the same as `type(self)`. It never changes and is the most derived class in the MRO (index 0).  It's `__thisclass__` that changes to reflect where in the MRO the new starting search point actually is.

I am aware of this. That's why I am (quite) sure that
what you need is

    mro = PyObject_CallMethodObjArgs(__self_class__, strmro, NULL);
    _NSCHECK(mro, NULL);

    index = PySequence_Index(mro, class_that_invoked_super);

With your code, `index` will always be 0 as for any class
it is the 0's element of its "mro".

d-maurer:
Jason Madden wrote at 2020-3-7 02:05 -0800:
> ...
>It's not particularly useful because the `super` object cannot access instance variables of `self` (only class descriptors/methods), nor (as a consequence) can it discover any interfaces that are `directlyProvidedBy` the `self` object. It has no MRO or type and `super` doesn't work on it, so you'd only be able to try this trick for exactly one level of hierarchy. The receiving factory would need to unwrap `__self__` in short order to do any of those things��and at that point, you're writing an adapter that's customized to know it's working with `super`, which destroys reusability: given `class Base` and `class Derived(Base)` the same adapter can't work for `super(Derived, self)` *and* `Base()`.


I am convinced.

src/zope/interface/_zope_interface_coptimizations.c
jamadden: >> I'm afraid you have that backwards. In `super(Class, self)` `__self_class__` is always the same as `type(self)`. It never changes and is the most derived class in the MRO (index 0). It's `__thisclass__` that changes to reflect where in the MRO the new starting search point actually is.

> I am aware of this. That's why I am (quite) sure that
> what you need is
```
> mro = PyObject_CallMethodObjArgs(__self_class__, strmro, NULL);
> _NSCHECK(mro, NULL);
>
> index = PySequence_Index(mro, class_that_invoked_super);
```

> With your code, `index` will always be 0 as for any class
> it is the 0's element of its "mro".

Essentially true. We're letting `super` do the harder work of figuring out exactly where in MRO we are. I thought that the class linearization order, which preserves local precedence order, would handle everything else correctly while minimizing our search time here.  But I forgot about diamonds!

Continuing the example from above it looks good:
```pycon
>>> from zope.interface.declarations import _next_super_class
>>> _next_super_class(sd) # super(D, d) means begin looking above D
<class '__main__.C'>
>>> _next_super_class(sc)
<class '__main__.B'>
>>> _next_super_class(sb)
<class '__main__.A'>
```

Introducing a diamond shows the inconsistency:
```pycon
>>> class BB(A):
...     pass
...
>>> class E(B, BB):
...     pass
...
>>> E.__mro__
(<class '__main__.E'>,
 <class '__main__.B'>,
 <class '__main__.BB'>,
 <class '__main__.A'>,
 <class 'object'>)
>>> e = E()
>>> se = super(E, e)
>>> sb = super(B, e)
>>> sbb = super(BB, e)
>>> _next_super_class(se)
<class '__main__.B'>
>>> _next_super_class(sb) # Should be BB
<class '__main__.A'>
>>> _next_super_class(sbb)
<class '__main__.A'>
```

Redefining as suggested is consistent. Thank you!
```pycon
>>> def _next_super_class(ob):
...     # When ``ob`` is an instance of ``super``, return
...     # the next class in the MRO that we should actually be
...     # looking at.
...     class_that_invoked_super = ob.__self_class__
...     mro = class_that_invoked_super.__mro__
...     next_class = mro[mro.index(ob.__thisclass__) + 1]
...     return next_class
…
>>> _next_super_class(se)
<class '__main__.B'>
>>> _next_super_class(sb)
<class '__main__.BB'>
>>> _next_super_class(sbb)
<class '__main__.A'>
```

jamadden:
CPython 3.5 is crashing on Travis/Linux, apparently hanging on Travis/Mac, and crashing locally for me on my Mac. 

The local crash occurs in `test_getMultiAdapter_hit_super (zope.interface.tests.test_registry.ComponentsTests)` and looks like an issue computing a hash code (?) when getting a key in a dictionary (compiled with `-g -O0`:

```bt
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x00007fff67191774 libsystem_malloc.dylib`large_malloc + 89
    frame #1: 0x00007fff67187357 libsystem_malloc.dylib`szone_malloc_should_clear + 235
    frame #2: 0x00007fff6718606a libsystem_malloc.dylib`malloc_zone_malloc + 104
    frame #3: 0x00007fff67185fe5 libsystem_malloc.dylib`malloc + 21
    frame #4: 0x000000010012f012 .Python`_PyObject_Alloc + 881
    frame #5: 0x0000000100120030 .Python`PyLong_FromDouble + 176
    frame #6: 0x00000001026f1f23 _zope_interface_coptimizations.cpython-35m-darwin.so`_lookup(self=0x000000010631cea8, required=0x0000000106322b08, provided=0x00000001060ff9a0, name=0x0000000100553ab0, default_=0x0000000000000000) at _zope_interface_coptimizations.c:1015:12
    frame #7: 0x00000001026f1b59 _zope_interface_coptimizations.cpython-35m-darwin.so`lookup_lookup(self=0x000000010631cea8, args=0x00000001063195e8, kwds=0x0000000000000000) at _zope_interface_coptimizations.c:1060:10
    frame #8: 0x000000010012c1f7 .Python`PyCFunction_Call + 107
```

where `_zope_interface_coptimizations.c:1015` is `result = PyDict_GetItem(cache, key);`

jamadden:
The crash in 3.5 turned out to be a memory bug in ``tuple(map(func, it))`` when `func` could return temporary objects. The fix is to replace `map(func, it)` with `[func(i) for i in it]`. This turns out to be faster too.

The temporary objects in question are the objects created for `providedBy(super(C, c))`. Unlike `providedBy(c)`, those aren't currently cached anywhere else (in an instance or class dict). But they'll wind up cached somewhat invisibly in adapter registries. They'll be found again (hashcodes and eq match up) so we're just doing some extra work computing the RO each time. I'm wondering if we should cache that in the class dict to avoid that.

Right now, I'm thinking leave it simple and revisit if it becomes a bottleneck.

jamadden:
> I am also worried that the use of tuple(map(func, it)) (over [func(i) for i in it]) could cause a crash. In my view, this indicates that something is deeply wrong and may hit people without the capabilities to find out why and how to work around it. Did the problem only affected Python 3.5 or other Python 3 versions as well (in other works, were the bug in Python 3.5 or somewhere in zope.interface).

Yeah, I was concerned about that too and kept looking into it. The crash only happened in CPython 3.5, and debugging showed that the call to `tuplefy(required)` resulted in the  dictionary stored in the `cache` C variable getting deallocated. Since `tuplefy` was iterating a lazy mapping that's calling arbitrary Python code, arbitrary side-effects are possible. In this case, it turns out that one such side-effect was eventually calling `components.changed(_)`, which drops the caches. Since `cache` contained a borrowed reference, its refcount went to 0 and it got deallocated. (This happened on every version of Python, but for some reason having to do with the `dict` implementation or memory management, it only crashed on CPython 3.5.)

The change to non-lazily iterate before calling into C is a sufficient fix *for this code* (although I can and will also eliminate the invoking of this side-effect in the first place). But this is a bug waiting to happen and the C code has been incorrect for a long time. The remainder of the fix is to not get the cache until we can be (more) sure there will be no more arbitrary side effects.

jamadden:
> Something is still inconsistent….Either tuple(providedBy(super(M1, d))) should not contain IB or tuple(providedBy(super(M2, d))) should contain IB (because M1 and M2 are symmetric with respect to B). Very likely, the latter should be the case, as super(M2, d) is "d viewed as B object" and B implements IB.

You're absolutely right. I've added that as a test case and corrected it.

src/zope/interface/adapter.py
```diff
@@ -30,6 +30,22 @@
...
+# ``tuple([t for t in range(10)])``      -> 610ns
+# ``tuple(t for t in range(10))``        -> 696ns
+# ``tuple(map(lambda t: t, range(10)))`` -> 881ns
+##
+# 2.7
+# ``tuple([t fon t in range(10)])``      -> 625ns
+# ``tuple(t for t in range(10))``        -> 665ns
+# ``tuple(map(lambda t: t, range(10)))`` -> 958ns
+#
+# All three have substantial variance.
```
d-maurer: The comment looks somewhat unmotivated at this place (though I know from the conversation in this PR why it is somewhere).

jamadden:
Thanks for the detailed reviews and attention to detail!

> Should we have a test "adapter_hook hits super" to verify that the "__self__" dereferencing works in this case, too? Maybe, also for multi adapters?

Both of those things are covered in the tests, but I will look at making them more explicit.

PR #188, opened 2020-03-17
Ensure Interface is the last item in the __sro__.

### LINKED ISSUE #8: Base class with interface interrupts adapter registered for a class
In https://bugs.launchpad.net/zope.interface/+bug/226112, @ctheune reports:

> See the attached patch for a failing test case.
> 
> (Patch applicable against zope.interface trunk)
> 
> We're not sure whether this is right or wrong but we feel the intended behaviour is what our test expects.

The patch adding a testcase: https://bugs.launchpad.net/zope.interface/+bug/226112/+attachment/290210/+files/weird-bug.patch

jamadden:
None of the elegant solutions mentioned in the issue worked out, so I had to brute force it.

Fixes #8

Plone's buildout.coredev tests show no new regressions.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -160,7 +160,7 @@
 
 - Adopt Python's standard `C3 resolution order
   <https://www.python.org/download/releases/2.3/mro/>`_ to compute the
-  ``__iro___`` and ``__sro___`` of interfaces, with tweaks to support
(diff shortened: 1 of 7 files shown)


jamadden:
I'll take it 😄 Thank you!

PR #189, opened 2020-03-17
Documentation clarifications.

### LINKED ISSUE #134: names() implementation contradicts documentation in Python 3
There's something I noticed while porting some code to Python 3. The implementation of names() in InterfaceClass doesn't return a sequence, as the documentation states, but a dict view.
To me this seems weird – have a function return two different types that don't represent the same way/idea of working with the data underneath it. 
How do we proceed with this?
I'd prefer to return a list, even if we don't take advantage of the new dict behaviour, because this would be a breaking change in the API.

https://github.com/zopefoundation/zope.interface/blob/3042b9fa7c37aa0d0b61efcdb0b1a436fef9bf59/src/zope/interface/interface.py#L389-L399

### LINKED ISSUE #136: Documentation bug: subscription delivery order
"https://zopeinterface.readthedocs.io/en/latest/" states in section "subscriptions":

> Subscribers defined for less general required interfaces are returned before subscribers for more general interfaces.

This contraticts the example preceding this quote where the subscription for `None` (the most general interface) is delivered before the subscription for  (the less general interface) `IRequire1`.

In addition, the documentation is not yet Python 3 compatible.

jamadden:
- docs/adapter.rst
  Subscriptions are returned from least to most specific, not the other way around; the docs were incorrect.

  Add additional examples, and use more verbose names in current examples, to clarify this. Fixes #136.
- interfaces.py
  names() and namesAndDescriptions() just return a collection, not a sequence. Fixes #134.

### DIFF
--- a/docs/adapter.rst
+++ b/docs/adapter.rst
@@ -24,27 +24,27 @@ Let's look at a simple example, using a single required specification:
   >>> from zope.interface.adapter import AdapterRegistry
   >>> import zope.interface
 
-  >>> class IRequire1(zope.interface.Interface):
+  >>> class IRequireBase(zope.interface.Interface):
   ...     pass
-  >>> class IProvide1(zope.interface.Interface):
+  >>> class IProvideBase(zope.interface.Interface):
   ...     pass
-  >>> class IProvide2(IProvide1):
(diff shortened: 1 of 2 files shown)


jamadden:
Thank you!

PR #191, opened 2020-03-18
Make Interface.getTaggedValue follow the __iro__.

### LINKED ISSUE #190: Interface.get/queryTaggedValue should respect __iro__, not just __bases__
Consider this inheritance tree:
```python
        O = Interface
        class F(O):
            taggedValue('tag', 'F')
            tag = Attribute('F')
        class E(O):
            taggedValue('tag', 'E')            
            tag = Attribute('E')
        class D(O):
            taggedValue('tag', 'D')            
            tag = Attribute('D')
        class C(D, F):
            taggedValue('tag', 'C')            
            tag = Attribute('C')
        class B(D, E):
            pass
        class A(B, C):
            pass
```

If we then look at the tagged value `"tag"` and the `tag`  Attribute of `A`, we get two different answers:

```pycon
>>> A['tag'].__name__
'C'
>>> A.getTaggedValue('tag')
'D'
```

This is because looking up tagged values only uses `__bases__`, whereas finding attributes follows the `__iro__`:
```pycon
>>> A.__iro__
(<InterfaceClass __main__.A>,
 <InterfaceClass __main__.B>,
 <InterfaceClass __main__.C>,
 <InterfaceClass __main__.D>,
 <InterfaceClass __main__.E>,
 <InterfaceClass __main__.F>,
 <InterfaceClass zope.interface.Interface>)
```

To implement this, much like there exists `InterfaceClass.direct(name)`, there should probably exist `InterfaceClass.queryDirectTaggedValue(name)`. 

Other `Direct` methods, like `getDirectTaggedValueTags` might be nice (though I don't have a direct use case for them right now).

jamadden:
Previously it manually walked up __bases__, meaning the answers could be inconsistent.

Fixes #190.

Also fixes several minor issues in the documentation, mostly cross-reference related.

### DIFF
--- a/docs/api/specifications.rst
+++ b/docs/api/specifications.rst
@@ -23,6 +23,7 @@ Specification objects implement the API defined by
    :member-order: bysource
 
 .. autoclass:: zope.interface.interface.Specification
+   :no-members:
 
 For example:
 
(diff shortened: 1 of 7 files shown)


docs/api/specifications.rst
```diff
@@ -172,7 +173,22 @@ first is that of an "element", which provides us a simple way to query
 for information generically (this is important because we'll see that
 ``IInterface`` implements this interface):
 
+..
+  IElement defines __doc__ to be an Attribute, so the docstring
+  in the class isn't used._
+
```
mgedmin: Ooh, what a nice corner case!

jamadden:
Thanks!

PR #198, opened 2020-03-30
Ensure that objects that implement no interfaces still have Interface in iro and sro

### LINKED ISSUE #197: Regression: Objects with no declared interfaces used to implement Interface
In 5.0.x, the `_empty` `_ImmutableDeclaration` object does not include `Interface`, whereas in 4.7 `_empty` did. This breaks looking up adapter registrations for `Interface` if the provided object does not otherwise implement any interfaces. This includes things like:

- Instances of direct subclasses of `object`;
- (Python 2) Old style classes (not their instances).

See https://github.com/zopefoundation/zope.traversing/issues/14 for a practical failure.

jamadden:
Fixes #197.

I've confirmed this fixes the zope.traversing tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,12 @@
 5.0.2 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Ensure that objects that implement no interfaces (such as direct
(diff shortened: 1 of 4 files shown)


jamadden:
Thanks!

PR #203, opened 2020-04-06
Make @implementer and classImplements not add redundant interfaces

jamadden:
The plone buildout.coredev tests all pass, with the exception of 8 tests from `plone.app.caching.tests.test_etags`. They all fail like this:

```
Error in test test_Roles_member (plone.app.caching.tests.test_etags.TestETags)
  File "//2.7/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "//plone.app.caching-2.0.4-py2.7.egg/plone/app/caching/tests/test_etags.py", line 144, in test_Roles_member
    provideAdapter(DummyPortalState, name=u'plone_portal_state')
  File "//zope.component-4.5-py2.7.egg/zope/component/globalregistry.py", line 73, in provideAdapter
    base.registerAdapter(factory, adapts, provides, name, event=False)
  File "//src/zope.interface/src/zope/interface/registry.py", line 305, in registerAdapter
    provided = _getAdapterProvided(factory)
  File "//src/zope.interface/src/zope/interface/registry.py", line 528, in _getAdapterProvided
    "The adapter factory doesn't implement a single interface "
TypeError: The adapter factory doesn't implement a single interface and no provided interface was specified.
```

That's with code like this:

```python
       @implementer(Interface)
       @adapter(DummyContext, Interface)
        class DummyPortalState(object):

            def __init__(self, context, request):
                pass

            def language(self):
                return 'en'

        provideAdapter(DummyPortalState, name=u'plone_portal_state')
```

Something is going wrong with `@implementer(Interface)`...

### DIFF
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -431,8 +431,13 @@ def classImplementsOnly(cls, *interfaces):
     in *interfaces*. This can be used to alter the interface resolution order.
     """
     spec = implementedBy(cls)
+    # Clear out everything inherited. It's important to
+    # also clear the bases right now so that we don't improperly discard
+    # interfaces that are already implemented by *old* bases that we're
+    # about to get rid of.
     spec.declared = ()
     spec.inherit = None
+    spec.__bases__ = ()
     _classImplements_ordered(spec, interfaces, ())
 
 
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -418,18 +418,21 @@ def implementedBy(cls): # pylint:disable=too-many-return-statements,too-many-bra
 
 
 def classImplementsOnly(cls, *interfaces):
-    """Declare the only interfaces implemented by instances of a class
+    """
+    Declare the only interfaces implemented by instances of a class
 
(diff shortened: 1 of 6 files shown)


jamadden:
Latest commits make plone's buildout.coredev green for me. Currently based on #202.

jamadden:
Rebased on master.

src/zope/interface/declarations.py
```diff
@@ -483,13 +500,30 @@ def classImplementsFirst(cls, iface):
...
+
     # eliminate duplicates
     new_declared = []
     seen = set()
-    for b in before + spec.declared + after:
-        if b not in seen:
-            new_declared.append(b)
-            seen.add(b)
+    for l in before, spec.declared, after:
+        for b in l:
```
mgedmin, line 523: Does pylint not complain about using lowercase-ell as a variable name?

jamadden: It probably would, but the check for valid names has to be disabled when working in zope repos because the false positives are overwhelming. (It's not the first one-letter temporary variable in this file.)

image

jamadden:
Thank you.

PR #205, opened 2020-04-07
The ImmutableDeclaration also has immutable _v_attrs.

### LINKED ISSUE #204: ImmutableDeclaration: Allows mutable _v_attrs
Some user code likes to assign to `_v_attrs` and/or write into the dictionary to piggyback on caching. Since ImmutableDeclaration provides no way to clear `_v_attrs` (it assumes it stays None), this can lead to incorrect results. 

This is not a regression in 5.0 but a pre-existing problem.

Proposal: Make `ImmutableDeclaration` use a property to always return a fresh dictionary for `_v_attrs`.

jamadden:
Fixes #204

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,12 @@
 
   See `issue 3 <https://github.com/zopefoundation/zope.interface/issues/3>`_.
 
+- Make the internal singleton object returned by APIs like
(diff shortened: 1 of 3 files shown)


jamadden:
Thank you.

PR #215, opened 2020-09-25
Make sure to call each invariant only once when validating invariants.

janjaapdriessen:
When an invariant is defined in an interface, it's found by `validateInvariants` in all interfaces inheriting from that interface.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,9 +12,11 @@
   that argument. See `issue 208
   <https://github.com/zopefoundation/zope.interface/issues/208>`_.
 
-- Make sure to call each invariant only once when validating invariants. 
-  Previously, invariants could be called multiple times because when an invariant is defined in an interface, it's found by in all interfaces inheriting from that interface.
-   See `pull request 215 <https://github.com/zopefoundation/zope.interface/pull/215/>`_.
+- Make sure to call each invariant only once when validating invariants.
+  Previously, invariants could be called multiple times because when an
+  invariant is defined in an interface, it's found by in all interfaces
+  inheriting from that interface.  See `pull request 215
+  <https://github.com/zopefoundation/zope.interface/pull/215/>`_.
 
 5.1.0 (2020-04-08)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,9 +12,9 @@
   that argument. See `issue 208
   <https://github.com/zopefoundation/zope.interface/issues/208>`_.
 
-- When an invariant is defined in an interface, it's found by
-  `validateInvariants` in all interfaces inheriting from that interface.
-  Make sure to call each invariant only once when validating invariants.
+- Make sure to call each invariant only once when validating invariants. 
+  Previously, invariants could be called multiple times because when an invariant is defined in an interface, it's found by in all interfaces inheriting from that interface.
+   See `pull request 215 <https://github.com/zopefoundation/zope.interface/pull/215/>`_.
 
 5.1.0 (2020-04-08)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@
   that argument. See `issue 208
   <https://github.com/zopefoundation/zope.interface/issues/208>`_.
 
+- When an invariant is defined in an interface, it's found by
+  `validateInvariants` in all interfaces inheriting from that interface.
+  Make sure to call each invariant only once when validating invariants.
 
 5.1.0 (2020-04-08)
 ==================
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -875,29 +875,21 @@ class InterfaceClass(_InterfaceClassBase):
     def queryDescriptionFor(self, name, default=None):
         return self.get(name, default)
 
-    def __validateInvariants(self, obj, errors, seen):
-        for call in self.queryTaggedValue('invariants', []):
-            if call in seen:
-                continue
-            seen.add(call)
-            try:
-                call(obj)
-            except Invalid as e:
-                if errors is None:
-                    raise
-                errors.append(e)
-        for base in self.__bases__:
-            try:
-                base.__validateInvariants(obj, errors, seen)
-            except Invalid:
-                if errors is None:
-                    raise
-        if errors:
-            raise Invalid(errors)
-
     def validateInvariants(self, obj, errors=None):
         """validate object to defined invariants."""
-        self.__validateInvariants(obj, errors, set())
+
+        for iface in self.__iro__:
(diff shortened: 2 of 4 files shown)


CHANGES.rst
```diff
@@ -12,6 +12,9 @@
   that argument. See `issue 208
   <https://github.com/zopefoundation/zope.interface/issues/208>`_.
 
+- When an invariant is defined in an interface, it's found by
+  `validateInvariants` in all interfaces inheriting from that interface.
+  Make sure to call each invariant only once when validating invariants.
```
jamadden, line 17:
```suggestion
- Make sure to call each invariant only once when validating invariants. 
  Previously, invariants could be called multiple times because when an invariant is defined in an interface, it's found by in all interfaces inheriting from that interface.
   See `pull request 215 <https://github.com/zopefoundation/zope.interface/pull/215/>`_.
```

We try to write change log entries in imperative voice, i.e., starting with a verb.

src/zope/interface/interface.py
```diff
@@ -875,9 +875,14 @@ def direct(self, name):
     def queryDescriptionFor(self, name, default=None):
         return self.get(name, default)
 
-    def validateInvariants(self, obj, errors=None):
+    def validateInvariants(self, obj, errors=None, seen=None):
```
jamadden, line 878: The implementation seems fine to me. Invariants might not be hashable, but that seems unlikely.

`validateInvariants` is a public method with a documented signature, though, and I'm not sure I like adding a public argument to that. The documentation and introspection by things like IDEs will disagree. I've seen this handled in two different ways. One is to name the argument with a leading underscore, indicating its for internal usage.
```suggestion
    def validateInvariants(self, obj, errors=None, _seen=None):
```

The other is to move the real work to an internal function:
```suggestion
    def validateInvariants(self, obj, errors=None):
        return self.__validateInvariants(obj, errors, set())
```
This has the advantage of not needing to check that `seen` is None at the start of the function. (We'd use `base.__validateInvariants(obj, errors, seen)` below too, of course.)

It's possible, (but not likely), that someone has subclassed `InterfaceClass` and overridden `validateInvariants` or used `@interfacemethod` to write their own `validateInvariants` method. The current implementation, as well as the first alternate form, could break such classes. I think the second alternate form (internal implementation) would be safe, but could wind up bypassing the custom methods. 

Thoughts?

jamadden (commented):
Thanks for this!

src/zope/interface/interface.py
janjaapdriessen: Thank you for your thorough explanation @jamadden. My first instinct would be to follow the `_seen` implementation, but moving the code to  a `__validateInvariants` [looks cleaner](https://github.com/zopefoundation/zope.interface/pull/215/commits/64fb805b3b2a6536c4874b4f0c3fbbbf4bea9e23), so I am okay with the second approach.

jamadden:
Thanks for considering that! I see your new implementation and I agree it looks nice and clean!

While reviewing it, I had another thought though: What if the body was changed from `self.queryTaggedValue()` to `self.queryDirectTaggedValue()`? That would be only change to the entirety of the original function. 

`queryTaggedValue` returns the first value seen when walking up the `__iro__`. That could result in the same value being seen multiple times if the tag is defined somewhere up the `__iro__` from the current object: the problem this PR is preventing.

`queryDirectTaggedValue`, OTOH, looks only and exactly for a value defined in this exact instance. That would eliminate the entire possibility of finding duplicates through the inheritance hierarchy. 

As this function proceeds to walk up `__bases__` it should ultimately always find everything included in `__iro__`….

But here's the problem with that: it might actually find them multiple times:

```pycon
>>> class I1(Interface):
...     pass
...
>>> class I2(I1):
...     pass
...
>>> class I11(I1): 
…     pass
…
>>> class I3(I11, I2):
...     pass
...
>>> I3.__iro__
(<InterfaceClass __main__.I3>,
 <InterfaceClass __main__.I11>,
 <InterfaceClass __main__.I2>,
 <InterfaceClass __main__.I1>,
 <InterfaceClass zope.interface.Interface>)
>>> def visit(s, pfx=''):
...     print(pfx, s)
...     for b in s.__bases__:
...         visit(b, pfx + ' ')
...
>>> visit(I3)
 <InterfaceClass __main__.I3>
  <InterfaceClass __main__.I11>
   <InterfaceClass __main__.I1>
    <InterfaceClass zope.interface.Interface>
  <InterfaceClass __main__.I2>
   <InterfaceClass __main__.I1>
    <InterfaceClass zope.interface.Interface>
``` 

Actually, thinking more about this, I find this method very confusing as it currently exists. It uses `__iro__` to find invariants, but it walks up `__bases__`. Since this is `InterfaceClass`, `__bases__` should only consist of other interfaces (not specifications or anything like that), like `__iro__`. One consequence of this oddity is that the method is `O(n^2)` due to the hidden loop in `queryTaggedValue`; also, recursion isn't especially fast in Python.

What if the method was made non-recursive using `__iro__` and `queryDirectTaggedValue` directly? 

```python
def validateInvariants(self, obj, errors=None):
    # pseudo-code
    for iface in self.__iro__: # Iterate starting with self
        for invariant in iface.queryDirectTaggedValue('invariants', ()):
            try:
                invariant(obj)
            except Invalid:
                 if errors is not None: errors.append(e)
                 else: raise
    if errors: raise Invalid(errors)
```

That seems a lot simpler and more direct. It eliminates the hashing of the PR. It still has the same issue with people that have tried to override `validateInvariants` as the PR, but in reality that's probably a non-issue.

`queryDirectTaggedValue` is relatively new, so that might explain why it wasn't written this way to start with. It was added specifically because using `__bases__` [can result in inconsistencies](https://github.com/zopefoundation/zope.interface/issues/190).

Thoughts? What have I missed?

CHANGES.rst
```diff
@@ -12,6 +12,9 @@
   that argument. See `issue 208
   <https://github.com/zopefoundation/zope.interface/issues/208>`_.
 
+- Make sure to call each invariant only once when validating invariants.
```
jamadden, line 15: I think the indentation is probably off here, causing [the failures of `twine check`](https://travis-ci.org/github/zopefoundation/zope.interface/jobs/730943537#L339). Sorry about that.

janjaapdriessen:
The implementation using `queryDirectTaggedValue` looks smooth, all tests pass, and the issue in my codebase is fixed. I am happy with this fix, thanks for you help @jamadden!

jamadden (approved):
LGTM! Thanks for being willing to iterate!

janjaapdriessen:
Would you mind making a pypi release so I can clean up loose ends at my end? Thanks!

jamadden:
Done! And thank you! I haven't benchmarked but for deep interface trees I expect this also results in a small performance boost too.

PR #217, opened 2020-09-28
C optimizations: Spec_clear and Spec_traverse need to include Spec->_bases

### LINKED ISSUE #216: C implementation of `@implementer` leaks objects
It appears that the C implementation of `@implementer` has a reference leak. This leads to a gradual increase in GC time, in a long-running application.

The following test script:

```python
#!/usr/bin/env python

import gc
from zope.interface.declarations import implementer
from zope.interface.interface import Interface

for i in range(20000):
    @implementer(Interface)
    class TestClass:
        pass

    if i % 1000 == 0:
        gc.collect()
        n = len(gc.get_objects())
        print("After %i cycles: %i objects" % (i, n))
```

produces the following output:

```
After 0 cycles: 8199 objects
After 1000 cycles: 45199 objects
After 2000 cycles: 82199 objects
After 3000 cycles: 119199 objects
After 4000 cycles: 156199 objects
After 5000 cycles: 193199 objects
After 6000 cycles: 230199 objects
After 7000 cycles: 267199 objects
After 8000 cycles: 304199 objects
After 9000 cycles: 341199 objects
After 10000 cycles: 378199 objects
After 11000 cycles: 415199 objects
After 12000 cycles: 452199 objects
After 13000 cycles: 489199 objects
After 14000 cycles: 526199 objects
After 15000 cycles: 563199 objects
After 16000 cycles: 600199 objects
After 17000 cycles: 637199 objects
After 18000 cycles: 674199 objects
After 19000 cycles: 711199 objects
```

In other words, each invocation of `@implementer` leads to new objects which cannot be garbage-collected.

Disabling the C implementations (by moving aside the `_zope_interface_coptimizations.cpython-36m-x86_64-linux-gnu.so` file) resolves the problem, as does downgrading to zope.interface 4.7.2.

jamadden:
Previously they did not, leading to a reference leak of a tuple.

Fixes #216

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,27 @@
   garbage collection times. See `issue 216
   <https://github.com/zopefoundation/zope.interface/issues/216>`_.
 
+  .. caution::
+
+     This leak could prevent interfaces used as the bases of
+     other interfaces from being garbage collected. Those interfaces
+     will now be collected.
+
+     One way in which this would manifest was that ``weakref.ref``
+     objects (and things built upon them, like
+     ``Weak[Key|Value]Dictionary``) would continue to have access to
+     the original object even if there were no other visible
+     references to Python and the original object *should* have been
+     collected. This could be especially problematic for the
+     ``WeakKeyDictionary`` when combined with dynamic or local
+     (created in the scope of a function) interfaces, since interfaces
+     are hashed based just on their name and module name. See the
+     linked issue for an example of a resulting ``KeyError``.
+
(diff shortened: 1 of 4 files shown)


jensens (approved):
After going through the Plone rabbit hole of testing this change in the context of dynamic interfaces and its testing code - I say go for merge!

jamadden:
Thank you for your help! 

I'll try to find somewhere in the docs and changelog to leave a bit about the problems encountered in plone.dexterity.

Issue #218, opened 2020-10-01
"Interface.__eq__" ignores the "Interface" semantics and renders dynamic interface definitions unreliable


d-maurer:
`Interface.__eq__` is only based on `__name__` and `__module__`. This results in a very surprising notion of equality as the following code demonstrates:
```pycon
>>> from zope.interface import Interface
>>> class I(Interface):
...   def f(): pass
... 
>>> I1 = I
>>> class I: pass
... 
>>> I1 == I
True
```
i.e. the interface `I1` is considered equal to the "normal" (non Interface) class `I`.

This renders local (i.e. in a function body) interface definitions unreliable as
https://github.com/zopefoundation/zope.interface/issues/216#issuecomment-701047858
(and following comments) demonstrates. The resulting errors can be non-deterministic and very difficult to analyze.

@jamadden explains in #216 that `__eq__` is defined as it is in order to allow interfaces to be used as keys in `BTrees`.  This may justify the definition **however** due to the potentially very difficult errors which may result from local interface definitions it **should be clearly documented** (which currently is not the case) with a big warning regarding the implication for local interface definitions (or overridden interface definitions as in the example above).

jensens:
Taking into account that hashing and equality are called very very often in the code, I am also +1 for good documentation but keeping the current state of implementation. Performance wise it works fine.

jamadden:
Because the example is unrealistic, ignores pickling/persistence, only results in problems in extremely rare corner cases, and has been the status quo for a decade, I do not consider this to be a significant problem. 

I consider [the existing documentation](https://zopeinterface.readthedocs.io/en/latest/api/specifications.html#equality-hashing-and-comparisons) adequate. I feel there is no need for "a big warning."

jensens:
I would not add a big warning, just I think it would be good to mention _explicit_ that (like in the context of mocking for testing):

```
class A1:
    def a1():
        class IA(Interface):
            pass

class A2:
    def a2():
        class IA(Interface):
            pass
```

will let you run into problems.

d-maurer:
The `Interface` (more generally `Specification`)  equality definition and its potential caveats are now covered by the (upcoming) documentation (--> #220).

PR #220, opened 2020-10-16
Document persistency and equality

d-maurer:
Documentation related to #218

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -1046,6 +1046,43 @@ functionality for particular interfaces.
    how to override functions in interface definitions and why, prior
    to Python 3.6, the zero-argument version of `super` cannot be used.
 
+
+Persistency and Equality
+========================
+
+An important design goal has been persistency support for interfaces.
+This allows different processes to use the same interface and
+to create persistent associations between (persistent) objects
+and interfaces. For example, an application can store an object
+together with its provided interfaces in a database; later, potentially
+in a different invocation, it can search the database for objects
+providing a given interface.
+
+To make an object persistent, it must have a persistent identifier,
+PID for short.
+It is this PID which identifies the object across different
+processes. In the context of interfaces, we want to support
+evolution similarly to the evolution of classes: even though
+we change the interface (e.g. add a method, change documentation), we
+still want to consider it as the same interface.
+
+Python's main persistency support comes from its ``pickle`` module.
+It uses as PID for code objects (such as classes and functions)
+the combinations of their ``__module__`` and ``name`` attributes.
+In analogy, the PID for an interface is defined
+as the combination of its
+``__module__`` and ``__name__`` giving persistency behavior similar
+to classes, including evolution support.
+
+Unlike classes, interfaces define their
+(runtime) equality in terms of their PID, i.e. two interfaces are
+considered equal if they have equal PIDs. Especially,
+two interfaces defined in the same module are considered equal
+if they have the same name, even if they are otherwise completely
+unrelated. In rare cases, this may lead to surprises - especially,
+when interfaces are defined dynamically (e.g. inside functions).
+
+
 .. [#create] The main reason we subclass ``Interface`` is to cause the
              Python class statement to create an interface, rather
              than a class.
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -1135,6 +1135,22 @@ old one?
    ...
    _pickle.PicklingError: Can't pickle <class 'Foo'>: it's not the same object as builtins.Foo
 
+A consequence of this is that only one object of the given name can be
+defined and pickled at a time. If we were to try to define a new ``Foo``
+class (remembering that normally the ``sys.modules[__name__].Foo =``
+line is automatic), we still cannot pickle the old one:
+
+.. doctest::
+
+   >>> orig_Foo = Foo
+   >>> class Foo(object):
+   ...    pass
+   >>> sys.modules[__name__].Foo = Foo # XXX, see below
+   >>> pickle.dumps(orig_Foo)
+   Traceback (most recent call last):
+   ...
+   _pickle.PicklingError: Can't pickle <class 'Foo'>: it's not the same object as builtins.Foo
+
 Or what if there simply is no global object?
 
 .. doctest::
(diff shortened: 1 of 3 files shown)


jensens (commented):
I think this nails it. Maybe a short example, addressing the last paragraph would help to get it out of grey theory, like

```
>>> def function1():
...     class IMyInterface(Interface): 
...         pass
...     return IMyInterface
... 
>>> def function2():
...     class IMyInterface(Interface): 
...         pass
...     return IMyInterface == function1()
... 
>>> function2()
True
```

d-maurer:
> I think this nails it. Maybe a short example, addressing the last paragraph would help to get it out of grey theory, like

@jensens I added an example. I did not use your example because it is very rare to compare interfaces directly in application code. Equality checks happen indirectly, for example inside `alsoProvides` (or when used as key in a `dict` as in your original problem observation). Therefore, I constructed an example involving `alsoProvides`.

jensens (approved):
Technical this explains the special behaviour well. That said, since I am not good enough in English to correct any grammar or alike.

docs/README.rst
```diff
@@ -1046,6 +1046,71 @@ functionality for particular interfaces.
    how to override functions in interface definitions and why, prior
    to Python 3.6, the zero-argument version of `super` cannot be used.
 
+
+Persistency and Equality
+========================
+
+An important design goal has been persistency support for interfaces.
```
jamadden, line 1053:
```suggestion
An important design goal has been persistence support for interfaces.
```

docs/README.rst
```diff
@@ -1046,6 +1046,71 @@ functionality for particular interfaces.
...
+This allows different processes to use the same interface and
+to create persistent associations between (persistent) objects
```
jamadden, line 1055:
```suggestion
to create persistent associations between objects
```

docs/README.rst
```diff
@@ -1046,6 +1046,71 @@ functionality for particular interfaces.
...
+and interfaces. For example, an application can store an object
+together with its provided interfaces in a database; later, potentially
+in a different invocation, it can search the database for objects
+providing a given interface.
+
+To make an object persistent, it must have a persistent identifier,
```
jamadden, line 1061: To avoid getting into the weeds too much, I would probably gloss over some of the details of persistence and ``pickle`` in particular and just talk about "stable identifiers" and "global objects".

docs/README.rst
```diff
@@ -1046,6 +1046,71 @@ functionality for particular interfaces.
...
+
+Python's main persistency support comes from its ``pickle`` module.
+It uses as PID for code objects (such as classes and functions)
+the combinations of their ``__module__`` and ``name`` attributes.
+In analogy, the PID for an interface is defined
+as the combination of its
+``__module__`` and ``__name__`` giving persistency behavior similar
+to classes, including evolution support.
+
+Unlike classes, interfaces define their
```
jamadden, line 1077: I would probably also add a comment about how, *like* classes, only one object with a given `__module__` and `__name__` can be pickled at a time. 

```pycon
>>> class Foo:
...     pass
>>> f1 = Foo
>>> class Foo:
...     pass
>>> pickle.dumps(f1)
PicklingError: Can't pickle <class '__main__.Foo'>: it's not the same object as __main__.Foo
```

 If you do this dynamically without watching the assigned names, none can be pickled.
```pycon
>>> f1 = type('Foo', (), {})
>>> f2 = type('Foo', (), {})
>>> pickle.dumps(f1)
PicklingError: Can't pickle <class '__main__.Foo'>: it's not the same object as __main__.Foo
>>> del Foo
>>> pickle.dumps(f2)
PicklingError: Can't pickle <class '__main__.Foo'>: attribute lookup Foo on __main__ failed
```

All of the above holds for interfaces.

Classes, however, are inconsistent, or at least inconvenient: They compare and hash by identity (and can't be sorted) but pickle by global identifier, meaning that the value of `==` can change. 
```pycon
>>> class Foo:
...     pass
>>> f1 = Foo
>>> pickled_f1 = pickle.dumps(Foo)
>>> class Foo:
...     pass
>>> f1 == Foo
False
>>> unpickled_f1 = pickle.loads(pickled_f1)
>>> unpickled_f1
<class '__main__.Foo'>
>>> unpickled_f1 == Foo # Surprise!
True
>>> unpickled_f1 == f1
False
```

This is not the case for interfaces:
```pycon
>>> class IFoo(Interface):
...     pass
>>> f1 = IFoo
>>> pickled_f1 = pickle.dumps(f1)
>>> class IFoo(Interface):
...     pass
>>> IFoo == f1
True
>>> unpickled_f1 = pickle.loads(pickled_f1)
>>> unpickled_f1
<InterfaceClass __main__.IFoo>
>>> unpickled_f1 == IFoo == f1
True
```

docs/README.rst
```diff
@@ -1046,6 +1046,71 @@ functionality for particular interfaces.
...
+    ...         pass
+    ...     class I2(Interface):
+    ...         pass
+    ...     alsoProvides(obj, I, I2)
+    ... 
+    >>> add_interfaces(obj)
+    >>> # we would expect that *obj* provides 3 interfaces at this place but
+    ... len(list(providedBy(obj)))
+    2
+
```
jamadden, line 1112: However, `obj` cannot actually be pickled. So this may not be the best example of persistence.
```pycon
>> from zope.interface import *
>>> class I(Interface):
...     pass
...
>>> class Obj(object):
...     pass
...
>>> obj = Obj()
>>> alsoProvides(obj, I)
None
>>> def add_interfaces(obj):
...     class I(Interface):
...         pass
...     class I2(Interface):
...         pass
...     alsoProvides(obj, I, I2)
...
>>> add_interfaces(obj)
None
>>> pickle.dumps(obj)
PicklingError: Can't pickle <InterfaceClass __main__.I>: it's not the same object as __main__.I
```

jamadden (commented):
Thank you for this!

I'm a +0 on it.

On the entirely positive side, I appreciate the explanation for some of the motivation for the comparison and hashing behaviour. 

On the 0 side, I feel like the example of the possible hazards for the unwary, would be better made a part of [the existing documentation](https://zopeinterface.readthedocs.io/en/latest/api/specifications.html#equality-hashing-and-comparisons) for this behaviour, where there's already one such example. This is especially the case because the example doesn't actually work with persistence.

I'm happy to prepare a PR like that, incorporating the motivation and new example from this PR. 

I'm not unhappy enough to ask that this one not be merged.

d-maurer:
Jason Madden wrote at 2020-10-21 04:34 -0700:
>@jamadden commented on this pull request.
> ...

Thank you for your detailed review!

I will soon (likely in the next week) incorparate your suggestions.

> ...
>On the 0 side, I feel like the example of the possible hazards for the unwary, would be better made a part of [the existing documentation](https://zopeinterface.readthedocs.io/en/latest/api/specifications.html#equality-hashing-and-comparisons) for this behaviour, where there's already one such example. This is especially the case because the example doesn't actually work with persistence.

I would not mind to refer to an example elsewhere in the documentation.

You are right that the example does not work with persistence.
In my view, this is not a drawback: surprising effects can
happen without any persistence involved.


>I'm happy to prepare a PR like that, incorporating the motivation and new example from this PR.

Great!

I would not mind, however, if you would directly improve the
existing PR rather than make you own one.

> ...
>Classes, however, are inconsistent, or at least inconvenient: They compare and hash by identity (and can't be sorted) but pickle by global identifier, meaning that the value of `==` can change.

I do understand this:

classes are first class citizens in Python; they are often
defined non globally and it is important that classes defined
at different "places" in the same module which happen to have
the same name are not compared equal.

Persistency is a secondary aspect for classes (a high percentage of
the Python users likely does not use `pickle`).
Consequently, `pickle` just insists that classes involved in
pickles must be defined globally (and the "right" one must be found
via the lookup).

> ...
>However, `obj` cannot actually be pickled. So this may not be the best example of persistence.

It was not designed as a persistency example but as an example
for the equality definition causing surprises.

jamadden:
> I would not mind, however, if you would directly improve the existing PR rather than make you own one.

I took you at your word and added some commits to this. (Whoops, at this writing realising I missed one, coming soon)

In an earlier comment I suggested that maybe it was good not to get into the weeds of pickle too much (that is, go into too much unnecessary detail). But as I was working on this, I found that actually giving some concrete details about pickle helped structure the discussion. Let me know how I can clarify or simplify.

d-maurer:
@jamadden Thank you for the great introduction to class and interface persistence and the examples demonstrating potential caveats of the `interface` equality defintion!

PR #222, opened 2020-11-05
Add support for Python 3.9.

icemac:


appveyor.yml
```diff
@@ -1,3 +1,4 @@
+image: Visual Studio 2019
```
mgedmin, line 1: I think this part [breaks Python 2.7 on Appveyor](https://ci.appveyor.com/project/mgedmin/zope-interface/builds/36147882/job/u74jepxgii99ix5k):

```
********************************************************************************
WARNING:
        An optional code optimization (C extension) could not be compiled.
        Optimizations for this package will not be available!
Microsoft Visual C++ 9.0 is required. Get it from http://aka.ms/vcpython27
********************************************************************************
```

mgedmin (commented):
LGTM, except for the part where Appveyor is broken on 2.7.

appveyor.yml
icemac: Let's try `Visual Studio 2015` (aka the default) even though it does not seem to have Python 3.9.

mgedmin: Our appveyor.yml has a powershell snippet that can install missing Python versions using the official installers from python.org.

mgedmin (approved):
I see one of the two Appveyor checks has passed.

appveyor.yml
icemac: Sorry, I always forget this.

PR #226, opened 2021-02-26
First pass at github actions.

### LINKED ISSUE #225: Move Mac and Linux builds to GitHub actions
Since Travis CI no longer works. This is probably a blocker for any other PRs.

Because of the C extensions this doesn't quite fit our standard model. 

I have some github actions YAML from other repos that deal with C extensions and building/uploading Mac/Linux[32-bit, 64-bit, aarch64] wheels that I can bring over as a starting point.

Someone with access to the `zope.wheelbuilder` account would need to add a repository secret for `TWINE_PASSWORD` (a token with scoped upload rights to `zope.interface`). (I could create a PyPI token from my personal account, but since we have a dedicated account I'd prefer not to.)

- [x] Environment that builds docs
- [x] 32-bit linux build/upload
- [x] 64-bit linux build/upload
- [x] aarch64 linux build/upload
- [x] coveralls
- [x] mac builds/upload
- [x] Pure-python tests
- [x] Compare other settings with existing "pure-python" configurations
- [x] Update badges
- [x] Update any docs references

jamadden:
Fixes #225. 

I know it's bulky, but the format itself requires a lot of repetition. (Custom actions that could save duplication are pretty limited still; I did [wind up using them to some good effect in RelStorage](https://github.com/zodb/relstorage/tree/master/.github) but they seemed like overkill here.) I know it's probably not ideal, but it also works and I'm hoping to get this project unblocked for further development. Still, any suggestions for improvements are of course welcome!

Someone will need to add a `TWINE_PASSWORD` GitHub repository secret that is a token for `zope.eggbuilder` to upload to zope.interface.

- Builds and uploads manylinux32/64/aarch64 wheels.
- Builds and uploads Mac wheels.
- Builds the docs.
- Runs tests with the C extension and without the C extension.
- Reports coverage to coveralls.
- Has the start of an environment to do linting.
- Removes .travis.yml


.github/workflows/tests.yml
```diff
@@ -0,0 +1,423 @@
...
+  # Because sharing code/steps is so hard, and because it can be
+  # extremely valuable to be able to get binary wheels without
+  # uploading to PyPI and even if there is some failure, (e.g., for
+  # other people to test/debug), the strategy is to divide the process
+  # into several different jobs. The first builds and saves the binary
+  # wheels. It has dependent jobs that download and install the wheel
+  # to run tests, build docs, and perform linting. Building the
+  # manylinux wheels is an independent set of jobs.
+  #
+  # This divisin is time-saving for projects that take awhile to
```
mgedmin, line 86:
```suggestion
  # This division is time-saving for projects that take awhile to
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,423 @@
...
+
+      - name: Build zope.interface
+        run: |
+          # Next, build the wheel *in place*. This helps ccache, and also lets us cache the configure
+          # output (pip install uses a random temporary directory, making this difficult).
+          python setup.py build_ext -i
+          python setup.py bdist_wheel
+          # Also install it, so that we get dependencies in the (pip) cache.
+          pip install -U coverage
+          pip install -U 'faulthandler; python_version == "2.7" and platform_python_implementation == "CPython"'
```
mgedmin, line 145: wow, neat, first time I see environment markers used on the command line; I didn't know this was allowed!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,423 @@
...
+          path: dist/*whl
+      - name: Publish package to PyPI (mac)
+        # We cannot 'uses: pypa/gh-action-pypi-publish@v1.4.1' because
+        # that's apparently a container action, and those don't run on
+        # the Mac.
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && startsWith(runner.os, 'Mac')
+        env:
+          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
+        run: |
+          twine upload --skip-existing dist/*
```
mgedmin, line 165: Why do the upload before running tests?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,423 @@
...
+        # twice on PyPy.
+        run: |
+          coverage run -p -m unittest discover -s src
+          PURE_PYTHON=1 coverage run -p -m unittest discover -s src
+          coverage combine
+          coverage report -i
+      - name: Submit to Coveralls
+        # This is a container action, which only runs on Linux.
+        if: ${{ startsWith(runner.os, 'Linux') }}
+        uses: AndreMiras/coveralls-python-action@develop
```
mgedmin, line 232: Is it a good idea to run bleeding edge action code instead of a fixed tag?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,423 @@
...
+          pip install -U pip
+          pip install -U wheel
+          pip install -U `ls dist/zope.interface-*`[test]
+      - name: Lint
+        # We only need to do this on one version, and it should be Python 3, because
+        # pylint has stopped updating for Python 2.
+        # TODO: Pick a linter and configuration and make this step right.
+        run: |
+          pip install -U pylint
+          # python -m pylint --limit-inference-results=1 --rcfile=.pylintrc zope.interface -f parseable -r n
```
mgedmin, line 337: Uh.

mgedmin (commented):
whee this is a lot, I don't think I can give it the attention it requires this weekend

some comments, because I did try to review it with what's left of my braim

.coveragerc
```diff
@@ -1,6 +1,8 @@
 [run]
 branch = True
 source = zope.interface
+# New in 5.0; required for the GHA coveralls submission.
+relative_files = True
```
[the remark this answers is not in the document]
jamadden: Hmm, I've needed it in every one, at least when using 5.0 and the action (instead of manually calling the `coveralls` command, which I found problematic). Perhaps one of those is different?

.manylinux-install.sh
```diff
@@ -2,6 +2,25 @@
...
+# Running inside docker
+# Set a cache directory for pip. This was
+# mounted to be the same as it is outside docker so it
+# can be persisted.
+export XDG_CACHE_HOME="/cache"
+# XXX: This works for macOS, where everything bind-mounted
+# is seen as owned by root in the container. But when the host is Linux
+# the actual UIDs come through to the container, triggering
+# pip to disable the cache when it detects that the owner doesn't match.
+# The below is an attempt to fix that, taken frob bcrypt. It seems to work on
```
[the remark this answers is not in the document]
jamadden: Thanks for catching my typos!  I'll batch those fixes locally.

.github/workflows/tests.yml
jamadden: It's pretty cool, isn't it? The pip command line is just like `requirements.txt` or `install_requires=` in that it works in terms of abstract "requirements,"  which include all that stuff about environment markers.

.github/workflows/tests.yml
jamadden: One basic reason, with a rationale.

Most importantly, making a release usually means using `zest.releaser`, which will upload the `.tar.gz` source file. Users that then attempt to just install `zope.interface` to get the latest version will have to have to be able to build C extensions while they wait for the new wheels to be built and uploaded, which can take awhile. I've had numerous issues opened (in other projects) about "hey I can't install your project" or "your newest release is missing wheels." Consequently, when a release is made, I try to optimize the build process to start getting wheels out to PyPI ASAP. People are both faster at noticing new releases, and also not pinning version requirements, than one might think.

I feel this is safe because at the time the release is made, CI was known to be green, and there are no code changes from the last CI run, only the textual things that `zest.releaser` does.

.coveragerc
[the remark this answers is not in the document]
jamadden: (The [action documentation](https://github.com/AndreMiras/coveralls-python-action) specifies this setting.)

.github/workflows/tests.yml
jamadden: Great point. That's what [the documentation](https://github.com/AndreMiras/coveralls-python-action) and other examples I've seen have done, so that's what I've been doing with this action too. 

We could try pinning to a tag; the newest tag is `v20201129`, making it three months old. If we did that, though, we'd probably also want to pin `coverage` to the version that was current at that time, wouldn't we? And I'm not sure we'd want to do that. Since the action also includes `coveralls`, we'd be pinning ourself to an older version of that, something we also don't usually do IIRC; I'd be concerned about the coveralls API changing.  In the past, IIRC, coveralls submission failures have been basically silent and hard to debug.

I'm not sure which is the worst of the two.

mgedmin: Well, if the docs say so, I suppose it's fine.

.coveragerc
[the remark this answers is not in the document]
mgedmin: This could be it, I've always been using `pip install coveralls && coveralls`.  (Last time I checked, which was some time ago, the official Coveralls action didn't support Python.)

.github/workflows/tests.yml
mgedmin: I guess what I find strange is that only MacOS builders do the upload early; manylinux builds do it at the very end.  I suppose Linux users don't find the lack of binary wheels an inconvenience.

jamadden: For i686/x86_64, the *entire* set of manylinux wheels is built and uploaded in less than two minutes. These jobs are both kicked off immediately (have no dependencies) and run in parallel with the other build steps. They don't do testing either. So the wait is minimal.

Things are a bit different for the aarch64 wheels, which *do* test first, and the Windows wheels (which also test first, but also run separately in parallel to the process here). Refactoring those could be done, but I would consider that a separate task.

jamadden:
Can this be a reasonable enough starting point to merge? If not, what blockers need addressed?

Reason being, I'd like to continue cargo-culting it around in order to unblock persistent as well (https://github.com/zopefoundation/persistent/issues/157).

jensens (approved):
Overall I think this is a good start. I am sure this topic will have some iterations in future, but that's OK.

jamadden:
Thanks!

PR #228, opened 2021-02-26
Let subclasses of BaseAdapterRegistry customize the data structures.

### LINKED ISSUE #224: BaseAdapterRegistry hardcodes non-Persistent data structures; can be painful in very large sites. Add hooks?
Problem Description
===============

`BaseAdapterRegistry` has a few attributes that contain its data, most importantly `_subscribers` and `_adapters`:

https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L76-L80

These start out as a plain, empty `list`. As items are added to the registry, `dict` objects are created inline and put into these lists. Here's part of `register`: https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L143-L146

The `subscribe` method has the same thing, with the addition of creating tuples: https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L245

In the end, you wind up with a collection of nested `dict`s contained in a `list`. The keys in the dicts are strings, interfaces, possibly classes, and other specifications (`implementedBy`, `providedBy`) objects. The leaf values can contain arbitrary user objects, for example, the utilities.

Because these aren't `Persistent` objects, when stored in a ZODB, the entire structure has to be written fresh each time its mutated (e.g., adding a single utility causes the entire data structure to get pickled). When loading a registry from ZODB, again, because these aren't persistent, the entire structure has to get unpickled and loaded into memory. If the leaf values themselves are persistent, while it doesn't require loading their entire state, this still results in the creation of ghost objects which go into the ZODB Connection's cache, possibly pushing out other useful objects.

In large sites, this can become quite painful. In one site that has several thousand persistent utilities of different kinds registered, we're seeing pickles of the adapter registry being over 1MB in size, and the time and memory taken to create ghosts of all the persistent utilities is also substantial.

The higher-level user of the registry, `Components`, has methods [`_init_registries`](https://github.com/zopefoundation/zope.interface/blob/master/src/zope/interface/registry.py#L179) and [`_init_registrations`](https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/registry.py#L185) that exist as hooks for subclasses. [`zope.component.persistentregistry.PersistentComponents`](https://github.com/zopefoundation/zope.component/blob/master/src/zope/component/persistentregistry.py#L46) overrides these methods to make some of the data structures persistent (but importantly, not `_subscribers` and `_adapters`).

Existing Workaround
===============

We have [a workaround for this](https://github.com/NextThought/nti.site/blob/4f8a9b351886b0bc271b4af5b335b3f14eac1785/src/nti/site/site.py#L284) based on overriding `def changed(self, originally_changed)`, examining the data structures, and converting parts of it (the biggest dicts) into persistent objects. This is very kludgy, though: it takes unnecessary execution time because we don't know actually what part of the data structure changed so we have to examine everything, and thus it may wakeup unneeded objects, and it doesn't fully traverse the whole tree.

Proposal
=======

I propose that the `BaseAdapterRegistry` grow similar hooks as `Components`. For example, instead of just doing `d = {}` when adding a registration, it would call `d = self._new_mapping()`. The default implementation would return `{}`, but `zope.component.persistentregistry.PersistentAdapterRegistry` would override that method to return `PersistentMapping`; we'd do something similar for creating new tuples/replacing existing tuples. Our subclass would override that to return `OOBTree`. (We'd lose the optimization of only doing this for big dicts, but it's probably worth it.)

The end result would be that loading a `PersistentComponents` object or its `PersistentAdapterRegistry` would only load a single top-level reference from the ZODB, and the rest of the data structure tree could be loaded on demand, as needed. That should help with loading time and memory pressure for large sites.

Because these new methods would only be called when mutating the registry, the fast-path of utility/adapter lookup should be unaffected (indeed, for things that aren't a subclass of `PersistentAdapterRegistry`, it would be totally unaffected). 

If there's interest, I'm happy to implement this.

Thoughts?

### LINKED ISSUE #227: BaseAdapterRegistry.subscribe/unsubscribe non-symmetric updating `_provided`, leaves orphan entries
While working on #224 I noticed that there were no tests for `BaseAdapterRegistry._provided` (which keeps a map of `{interface: reference count}` of interfaces it knows about in order to add them as "extendors" to the AdapterLookup classes).

The `register` method increments this count and stores it:
https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L154-L157
As does the `subscribe` method:
https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L247-L251

The inverse of `register`, `unregister`, decrements this count and either stores it, or, if it hits 0, removes it:
https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L219-L224

However, the inverse of `subscribe`, `unsubscribe`, appears to have a bug. It decrements the count, but only if it hits zero is it removed. It never stores it, which means it can never hit 0 just by unsubscribing.
https://github.com/zopefoundation/zope.interface/blob/bf0de08af2d53261d7eb0ce1e31f574f93524ef6/src/zope/interface/adapter.py#L305-L309

Making `unsubscribe` store the decremented value doesn't break any tests (and fixes the tests I was adding).  Because the high-level class `Components` automatically both registers and subscribes utilities, it's fairly easy to end up with orphaned entries in `_provided`. 

Here, I'll define an interface and class and register it both as the default (unnamed) and a named utility for the same interface.

```pycon
>>> from zope.interface import Interface, implementer
>>> from zope.interface.registry import Components
>>> class IFoo(Interface):
...     pass
...
>>> @implementer(IFoo)
... class Foo:
...     pass
...
>>> foo = Foo()
>>> foo2 = Foo()
>>> comp = Components()
>>> comp.registerUtility(foo)
None
>>> comp.registerUtility(foo2, name='bar')
None
```

Now, because `IFoo` was both provided and subscribed twice, it has a refcount of 4, and the two objects show up once each in `_adapters` and `_subscribers` of the adapter registry:

```pycon
>>> comp.utilities._provided
{<InterfaceClass __main__.IFoo>: 4}
>>> comp.utilities._adapters
[{<InterfaceClass __main__.IFoo>: {'': <__main__.Foo object at 0x10a09dfa0>,
                                   'bar': <__main__.Foo object at 0x10a08e140>}}]
>>> comp.utilities._subscribers
[{<InterfaceClass __main__.IFoo>: {'': (<__main__.Foo object at 0x10a09dfa0>,
                                        <__main__.Foo object at 0x10a08e140>)}}]
```

Now we can unregister both utilities, and their entries will be gone from `_adapters` and `_subscribers`:

```pycon
>>> comp.unregisterUtility(foo)
True
>>> comp.unregisterUtility(foo2, name='bar')
True
>>> comp.utilities._adapters, comp.utilities._subscribers
([], [])
```

But `IFoo` is still in `_provided` because of this issue:

```pycon
>>> comp.utilities._provided
{<InterfaceClass __main__.IFoo>: 2}
```

Because of the way `unsubscribe` calculates the delta (removing all equal objects) if I register the same identical object (which is if course equal to itself) as both the named and unnamed utility, this appears to work. That is, `_provided` has no remaining references to `IFoo`.

When using non-equal objects, this only works (empty `_provided`) if I make `unsubscribe` store the decremented value. The equal object case still works then.

jamadden:
Add extensive tests for this. Fixes #224.

Also adds test for, and fixes #227

Based on #226 so that CI runs. The first two commits aren't relevant to this PR.

I'd like to add a migration method to help existing persistent utilities convert to new data structures, but before that's done I'll open this as a draft so the general approach can be looked at.

### DIFF
--- a/src/zope/interface/adapter.py
+++ b/src/zope/interface/adapter.py
@@ -13,6 +13,7 @@
 ##############################################################################
 """Adapter management
 """
+import itertools
 import weakref
 
 from zope.interface import implementer
@@ -30,6 +31,7 @@ __all__ = [
     'VerifyingAdapterRegistry',
 ]
 
+# In the CPython implementation,
 # ``tuple`` and ``list`` cooperate so that ``tuple([some list])``
 # directly allocates and iterates at the C level without using a
 # Python iterator. That's not the case for
@@ -46,6 +48,19 @@ __all__ = [
 # ``tuple(map(lambda t: t, range(10)))`` -> 958ns
 #
 # All three have substantial variance.
+##
+# On PyPy, this is also the best option.
+##
+# PyPy 2.7.18-7.3.3
+# ``tuple([t fon t in range(10)])``      -> 128ns
+# ``tuple(t for t in range(10))``        -> 175ns
+# ``tuple(map(lambda t: t, range(10)))`` -> 153ns
+##
+# PyPy 3.7.9 7.3.3-beta
+# ``tuple([t fon t in range(10)])``      ->  82ns
+# ``tuple(t for t in range(10))``        -> 177ns
+# ``tuple(map(lambda t: t, range(10)))`` -> 168ns
+#
 
 class BaseAdapterRegistry(object):
     """
@@ -54,41 +69,54 @@ class BaseAdapterRegistry(object):
 
     Subclasses can set the following attributes to control how the data
     is stored; in particular, these hooks can be helpful for ZODB
-    persistence:
+    persistence. They can be class attributes that are the named (or similar) type, or
+    they can be methods that act as a constructor for an object that behaves
+    like the types defined here; this object will not assume that they are type
+    objects, but subclasses are free to do so:
 
     _sequenceType = list
-      This is the type used for our two top-level "byorder" sequences.
+      This is the type used for our two mutable top-level "byorder" sequences.
+      Must support mutation operations like ``append()`` and ``del seq[index]``.
       These are usually small (< 10). Although at least one of them is
       accessed when performing lookups or queries on this object, the other
       is untouched. In many common scenarios, both are only required when
       mutating registrations and subscriptions (like what
       :meth:`zope.interface.interfaces.IComponents.registerUtility` does).
-      This use pattern makes it an ideal candidate to be a ``PersistentList``.
+      This use pattern makes it an ideal candidate to be a
+      :class:`~persistent.list.PersistentList`.
     _leafSequenceType = tuple
       This is the type used for the leaf sequences of subscribers.
       It could be set to a ``PersistentList`` to avoid many unnecessary data
-      loads when subscribers aren't being used. See :meth:`_addValueToLeaf`
-      and :meth:`removeValueFromLeaf` for two methods you'll want to override.
+      loads when subscribers aren't being used. Mutation operations are directed
+      through :meth:`_addValueToLeaf` and :meth:`_removeValueFromLeaf`; if you use
+      a mutable type, you'll need to override those.
     _mappingType = dict
-      This is the type used for the keyed mappings. A ``PersistentMapping``
+      This is the mutable mapping type used for the keyed mappings.
+      A :class:`~persistent.mapping.PersistentMapping`
       could be used to help reduce the number of data loads when the registry is large
       and parts of it are rarely used. Further reductions in data loads can come from
-      using a ``OOBTree``, but care is required to be sure that all required/provided
+      using a :class:`~BTrees.OOBTree.OOBTree`, but care is required
+      to be sure that all required/provided
       values are fully ordered (e.g., no required or provided values that are classes
       can be used).
     _providedType = dict
-      This is the type used for the ``_provided`` mapping.
+      This is the mutable mapping type used for the ``_provided`` mapping.
       This is separate from the generic mapping type because the values
       are always integers, so one might choose to use a more optimized data
-      structure such as a ``OIBTree``. The same caveats apply as for ``_mappingType``.
+      structure such as a :class:`~BTrees.OIBTree.OIBTree`.
+      The same caveats regarding key types
+      apply as for ``_mappingType``.
 
     It is possible to also set these on an instance, but because of the need to
-    potentially also override :meth:`addValueToLeaf` and :meth:`removeValueFromLeaf`,
+    potentially also override :meth:`_addValueToLeaf` and :meth:`_removeValueFromLeaf`,
     this may be less useful in a persistent scenario; using a subclass is recommended.
 
     .. versionchanged:: 5.3.0
-        Add support for customizing the way data
+        Add support for customizing the way internal data
         structures are created.
+    .. versionchanged:: 5.3.0
+        Add methods :meth:`rebuild`, :meth:`allRegistrations`
+        and :meth:`allSubscriptions`.
     """
 
     # List of methods copied from lookup sub-objects:
@@ -170,7 +198,7 @@ class BaseAdapterRegistry(object):
     # Hooks for subclasses to define the types of objects used in
     # our data structures.
     # These have to be documented in the docstring, instead of local
-    # comments, because autodoc ignores the comment and just writes
+    # comments, because Sphinx autodoc ignores the comment and just writes
     # "alias of list"
     _sequenceType = list
     _leafSequenceType = tuple
@@ -182,14 +210,18 @@ class BaseAdapterRegistry(object):
         Add the value *new_item* to the *existing_leaf_sequence*, which may
         be ``None``.
 
-        If *existing_leaf_sequence* is not *None*, it will be an instance
-        of `_leafSequenceType`.
-
-        This method returns the new value to be stored. It may mutate the
-        sequence in place if it was not None and the type is mutable, but
-        it must also return it.
-
         Subclasses that redefine `_leafSequenceType` should override this method.
+
+        :param existing_leaf_sequence:
+            If *existing_leaf_sequence* is not *None*, it will be an instance
+            of `_leafSequenceType`. (Unless the object has been unpickled
+            from an old pickle and the class definition has changed, in which case
+            it may be an instance of a previous definition, commonly a `tuple`.)
+
+        :return:
+           This method returns the new value to be stored. It may mutate the
+           sequence in place if it was not ``None`` and the type is mutable, but
+           it must also return it.
 
         .. versionadded:: 5.3.0
         """
@@ -199,15 +231,24 @@ class BaseAdapterRegistry(object):
 
     def _removeValueFromLeaf(self, existing_leaf_sequence, to_remove):
         """
-        Remove the item *to_remove* from the (non-None, non-empty) *existing_leaf_sequence*
-        and return the mutated sequence.
+        Remove the item *to_remove* from the (non-``None``, non-empty)
+        *existing_leaf_sequence* and return the mutated sequence.
 
-        If there is more than one item that is equal to *to_remove* they must all be
-        removed.
+        Subclasses that redefine `_leafSequenceType` should override
+        this method.
 
-        May mutate in place or return a new object.
+        If there is more than one item that is equal to *to_remove*
+        they must all be removed.
 
-        Subclasses that redefine `_leafSequenceType` should override this method.
+        :param existing_leaf_sequence:
+           As for `_addValueToLeaf`, probably an instance of
+           `_leafSequenceType` but possibly an older type; never `None`.
+        :return:
+           A version of *existing_leaf_sequence* with all items equal to
+           *to_remove* removed. Must not return `None`. However,
+           returning an empty
+           object, even of another type such as the empty tuple, ``()`` is
+           explicitly allowed; such an object will never be stored.
 
         .. versionadded:: 5.3.0
         """
@@ -423,13 +464,16 @@ class BaseAdapterRegistry(object):
             return  # pragma: no cover
         len_old = len(old)
         if value is None:
-            # Removing everything; note that the type of ``new`` won't match
-            # the ``_leafSequenceType``, but that's OK because we're about
-            # to delete the entire entry anyway.
+            # Removing everything; note that the type of ``new`` won't
+            # necessarily match the ``_leafSequenceType``, but that's
+            # OK because we're about to delete the entire entry
+            # anyway.
             new = ()
         else:
             new = self._removeValueFromLeaf(old, value)
-        # new may have been mutated in place, so we cannot compare it to old
+        # ``new`` may be the same object as ``old``, just mutated in place,
+        # so we cannot compare it to ``old`` to check for changes. Remove
+        # our reference to it now to avoid trying to do so below.
         del old
 
         if len(new) == len_old:
@@ -491,13 +535,12 @@ class BaseAdapterRegistry(object):
             # The generator doesn't actually start running until we
             # ask for its next(), by which time the attributes will change
             # unless we do so before calling __init__.
-            from itertools import chain
             try:
                 first = next(it)
             except StopIteration:
                 return iter(())
 
-            return chain((first,), it)
+            return itertools.chain((first,), it)
 
         registrations = buffer(registrations)
         subscriptions = buffer(subscriptions)
--- a/src/zope/interface/adapter.py
+++ b/src/zope/interface/adapter.py
@@ -48,6 +48,48 @@ __all__ = [
 # All three have substantial variance.
 
 class BaseAdapterRegistry(object):
+    """
+    A basic implementation of the data storage and algorithms required
+    for a :class:`zope.interface.interfaces.IAdapterRegistry`.
+
+    Subclasses can set the following attributes to control how the data
+    is stored; in particular, these hooks can be helpful for ZODB
+    persistence:
+
+    _sequenceType = list
+      This is the type used for our two top-level "byorder" sequences.
+      These are usually small (< 10). Although at least one of them is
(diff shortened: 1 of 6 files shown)


jamadden:
I've added the migration method. It turns out to be useful to fix persistent registries that are suffering from the refcount issue described in #227 as well.

The next step would be to update `zope.component.persistentregistry.PersistentAdapterRegistry` to use `PersistentList` and `PersistentMapping`, as appropriate, for these data types. It wouldn't automatically convert anything, except when values are added/removed to the leaf of the subscribers tree; then, since we're writing anyway, it ought to  convert from tuples to `PersistentList` as needed.

jamadden:
A question about the three methods I added (`allRegistrations`, `allSubscriptions`, `rebuild`): Should those go in an interface, or be left as "implementation specific" in this class hierarchy?

jamadden:
Are there any interested reviewers? I'm really hoping to be able to release this feature soon as large registries can become a considerable pain point.

jensens:
I would like to review, but at the moment my time is a bit limited. If you give me 2 more days?

jamadden:
Of course, certainly! That wasn't meant as a threat to merge it unreviewed. Just as a request for reviews, with context.

src/zope/interface/adapter.py
```diff
@@ -119,6 +167,52 @@ def _createLookup(self):
...
+        sequence in place if it was not None and the type is mutable, but
+        it must also return it.
+
+        Subclasses that redefine `_leafSequenceType` should override this method.
+
+        .. versionadded:: 5.3.0
+        """
+        if existing_leaf_sequence is None:
+            return (new_item,)
+        return existing_leaf_sequence + (new_item,)
```
ale-rt, line 198: I do not know if performance is an issue here, but concatenating tuples generates a new object which might have some impact on performance

src/zope/interface/adapter.py
```diff
@@ -119,6 +167,52 @@ def _createLookup(self):
...
+        If there is more than one item that is equal to *to_remove* they must all be
+        removed.
+
+        May mutate in place or return a new object.
+
+        Subclasses that redefine `_leafSequenceType` should override this method.
+
+        .. versionadded:: 5.3.0
+        """
+        return tuple([v for v in existing_leaf_sequence if v != to_remove])
```
ale-rt, line 214: Removing the square parentheses should grant better performances because you will not create a temporary `list` instance that will be immediately discarded:
```
return tuple(v for v in existing_leaf_sequence if v != to_remove)
```

Maybe even better is:
```
return tuple(filter(lambda value: value !=  to_remove, existing_leaf_sequence)
```
but I have no benchmarks about this.

As before, having mutable objects here might have been better.

src/zope/interface/adapter.py
```diff
@@ -177,6 +271,44 @@ def registered(self, required, provided, name=u''):
...
+        Yields tuples ``(required, provided, name, value)`` for all
+        the registrations that this object holds.
+
+        These tuples could be passed as the arguments to the
+        :meth:`register` method on another adapter registry to
+        duplicate the registrations this object holds.
+
+        .. versionadded:: 5.3.0
+        """
+        for t in self._all_entries(self._adapters):
```
ale-rt, line 309: If this is Python 3 only it might be written as:
```
yield from self._all_entries(self._adapters)
```

src/zope/interface/adapter.py
```diff
@@ -303,13 +456,61 @@ def unsubscribe(self, required, provided, value=None):
...
+
+        # Grab the iterators, we're about to discard their data.
+        registrations = self.allRegistrations()
+        subscriptions = self.allSubscriptions()
+
+        def buffer(it):
+            # The generator doesn't actually start running until we
+            # ask for its next(), by which time the attributes will change
+            # unless we do so before calling __init__.
+            from itertools import chain
```
ale-rt, line 494: should this import go to the top?

src/zope/interface/adapter.py
```diff
@@ -303,13 +456,61 @@ def unsubscribe(self, required, provided, value=None):
...
+            try:
+                first = next(it)
+            except StopIteration:
+                return iter(())
+
+            return chain((first,), it)
```
ale-rt, line 500: Clever!

ale-rt (commented):
Smart work! I just added some small comments you can safely ignore

src/zope/interface/adapter.py
jamadden: True! But: 

(a) that's what the existing code did, so there shouldn't be a regression (other than the slight cost of the added method call); and 

(b) even if registering new utilities/adapters is on a critical fast path of an application (seems doubtful to me, but assumptions like that can obviously be wrong!), this is probably the least part of that, when considering actually creating the component, testing whether it needs to be added to the subscribers sequence, emitting the event about the registration, etc.

src/zope/interface/adapter.py
jamadden: > Removing the square parentheses should grant better performances...but I have no benchmarks about this.

Luckily, there are benchmarks about this very thing at the top of the file!

https://github.com/zopefoundation/zope.interface/blob/c28c205c91e73d3ca5e195bb0392d10046e13be0/src/zope/interface/adapter.py#L33-L47

I tested all three earlier when the code migrated off of the `map` idiom.

> As before, having mutable objects here might have been better.

Perhaps, but because there are many existing pickles with this structure, there's going to have to be a migration process. I'll address that in `zope.component.persistentregistery`

src/zope/interface/adapter.py
jamadden: > If this is Python 3 only

It is not. The core libraries (zope.toolkit, perhaps, plus Acquisition?) are still supporting Python 2.7.

src/zope/interface/adapter.py
jamadden: I was trying to avoid importing it unless it was needed, since this function will rarely be called. But I hadn't actually checked to see if that made a difference at all.

It turns out that `import pkg_resources` (which happens in `zope/__init__.py`) imports `itertools` always. Even when that can go away (when/if we go to pure Python-3-only namespace packages), somewhere in the startup process of Python 3.6 and above `itertools` is imported before user code even executes.

So it's essentially a free import and can be put at module scope with no penalty.

jamadden:
Thanks for the review @ale-rt !

src/zope/interface/adapter.py
ale-rt: Wow! I am amazed :)
https://github.com/zopefoundation/zope.interface/blob/c28c205c91e73d3ca5e195bb0392d10046e13be0/src/zope/interface/adapter.py#L33-L35
TIL, thanks!

jamadden: For the curious: `tuple(some_object)` calls [`PySequence_Tuple(some_object)`](https://github.com/python/cpython/blob/2fe408497e6838b6fb761e42e8f8ffc70bd6f3b1/Objects/abstract.c#L1908); if the `some_object` argument is exactly a `list`, we wind up in [`PyList_AsTuple(some_object)`](https://github.com/python/cpython/blob/2fe408497e6838b6fb761e42e8f8ffc70bd6f3b1/Objects/listobject.c#L2485), which in turn passes a direct reference to the list's C-level `PyObject*` array back to [`_PyTuple_FromArray`](https://github.com/python/cpython/blob/2fe408497e6838b6fb761e42e8f8ffc70bd6f3b1/Objects/tupleobject.c#L414). That function just does some C pointer walking to initialize the new new tuple's C array.

Every other path through `PySequence_Tuple` (except for the trivial case of `tuple(other_tuple)` resort to the much more expensive case of using generic `iter(some_object)` and `next()` calls to populate the tuple.

src/zope/interface/adapter.py
```diff
@@ -177,6 +312,44 @@ def registered(self, required, provided, name=u''):
...
+    @classmethod
+    def _allKeys(cls, components, i, parent_k=()):
+        if i == 0:
+            for k, v in components.items():
+                yield parent_k + (k,), v
+        else:
+            for k, v in components.items():
+                new_parent_k = parent_k + (k,)
+                for x, y in cls._allKeys(v, i - 1, new_parent_k):
+                    yield x, y
```
jensens, line 324: I can not estimate good how many component.items() there are. But from my gut feeling I guess enough to better first store `cls._allKeys` in a local variable and avoid one load at byte-code level. Like:
```suggestion
        else:
            allKeys = cls._allKeys
            for k, v in components.items():
                new_parent_k = parent_k + (k,)
                for x, y in allKeys(v, i - 1, new_parent_k):
                    yield x, y
```

src/zope/interface/adapter.py
```diff
@@ -177,6 +312,44 @@ def registered(self, required, provided, name=u''):
...
+
+    def _all_entries(self, byorder):
+        # Locally reference the `byorder` data; it might be replaced while
+        # this method is running.
+        for i, components in enumerate(byorder):
+            # We will have *i* levels of dictionaries to go before
+            # we get to the leaf.
+            for key, value in self._allKeys(components, i + 1):
```
jensens, line 332: Same here?

```suggestion
        # this method is running.
        allKeys =  self._allKeys
        for i, components in enumerate(byorder):
            # We will have *i* levels of dictionaries to go before
            # we get to the leaf.
            for key, value in allKeys(components, i + 1):
```

jensens (commented):
A great new feature. Overall first class code! See my comments for two performance details.

src/zope/interface/adapter.py
jamadden: I appreciate your concern for performance!

Here, because of the recursion and use of generators, all of which are implemented in Python and spread across a few different functions, plus the fact that there's lots of tuple creation and unpacking going on, I strongly suspect our time won't be dominated by that attribute lookup. 

But! I haven't benchmarked! I haven't benchmarked because I don't know what a performance goal for this code should be. The only use-case I have for it now is in rebuilding the datastructures with the `rebuild()` method, and that's not particularly performance sensitive (again, in my use-case) — it's a migration-type activity. So I was more concerned about code clarity/maintainability, and secondarily (because of the recursion) trying to keep the stack frames small. 

Perhaps, if need be, we can revisit when someone comes up with different use cases?

jensens: My  thinking here: It's still readable, it does not hurt, it is fast from the beginning.

jamadden: Given this fixture:
```python
from zope.interface.adapter import AdapterRegistry
import pprint

reg = AdapterRegistry()
for i in range(10000):
    iface = type(Interface)(
        'I' + str(i),
        (Interface,),
        {}
    )
    reg.subscribe([object()], iface, i)
```

Looping across `reg.allSubscriptions()` with a simple `for _ in reg.allSubscriptions(): pass` calls `reg._allKeys` 16,221,622 times.

With the code as written, IPython's `%timeit` says this loop takes:
```
15.6 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Making the change suggested here results in:
```
15.8 ms ± 278 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

So, no appreciable benefit. The bottlenecks are elsewhere.

jensens: Indeed fast enough.

jensens (approved):

src/zope/interface/adapter.py
jamadden: I was testing on CPython 3.8. When I saw that the second version actually got a bit slower, I remembered that starting in Python 3.7 there's the new bytecode `LOAD_METHOD` and `CALL_METHOD` that optimize method calling and avoid creating temporary method objects. So this form of optimization can actually turn out to be a pessimazation on those versions!

Given:
```python
In [1]: class C:
   ...:     def meth(self):
   ...:         pass
   ...:     def call_meth_direct(self):
   ...:         self.meth()
   ...:     def call_meth_indirect(self):
   ...:         m = self.meth
   ...:         m()
   …:
```

The direct call results in 10 bytecodes:
```pycon
In [3]: dis.dis(C.call_meth_direct)
  5           0 LOAD_FAST                0 (self)
              2 LOAD_METHOD              0 (meth)
              4 CALL_METHOD              0
              6 POP_TOP
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE
```

While the version that stores it in a variable first misses the optimization and is bigger:
```pycon
In [5]: dis.dis(C.call_meth_indirect)
  7           0 LOAD_FAST                0 (self)
              2 LOAD_ATTR                0 (meth)
              4 STORE_FAST               1 (m)

  8           6 LOAD_FAST                1 (m)
              8 CALL_FUNCTION            0
             10 POP_TOP
             12 LOAD_CONST               0 (None)
             14 RETURN_VALUE
```

So what about Python 2.7, where the optimization isn't present? I get 14.7ms as-is, and between 14.3ms and 15.9ms with the modified version on two subsequent runs (IPython on 2.7 doesn't report standard deviation, just the best time).

jamadden:
Thanks for reviewing @jensens and @ale-rt !

jamadden:
I'd like to tackle #230, and if it's easy, #229 then make a release.

PR #231, opened 2021-03-15
Add IAdapterRegistry.subscribed and Components.rebuildUtilityRegistryFromLocalCache

### LINKED ISSUE #230: Add AdapterRegistry.subscribed()
This would be analogous to `AdapterRegistry.registered()`.

My particular use case for this for consistency checking and error correction. When you call `Components.registerUtility()` , the utility is supposed to wind up in three places:

- Components._utility_registrations
- Components.adapters._adapters
- Components.adapters._subscribers

But we had a situation where a bug (in a subclass) caused only the first and third of those (probably!) to contain the utility. Correcting `_adapters` was easy, because we could check with `registered()` whether it was present or not. Correcting `_subscribers` required implementing a `subscribed()` method.

If we can get a `subscribed()` method, I could also add the error-detecting/correcting method we wrote to `Components`, if desired.

jamadden:
I'm very open to improved naming suggestions for `rebuildUtilityRegistryFromLocalCache`.

Fixes #230

### DIFF
--- a/src/zope/interface/adapter.py
+++ b/src/zope/interface/adapter.py
@@ -430,6 +430,26 @@ class BaseAdapterRegistry(object):
 
         self.changed(self)
 
+    def subscribed(self, required, provided, subscriber):
+        required = tuple([_convert_None_to_Interface(r) for r in required])
+        name = u''
+        order = len(required)
+        byorder = self._subscribers
+        if len(byorder) <= order:
+            return None
+
+        components = byorder[order]
+        key = required + (provided,)
+
+        for k in key:
+            d = components.get(k)
+            if d is None:
+                return None
+            components = d
+
+        subscribers = components.get(name, ())
+        return subscriber if subscriber in subscribers else None
+
     def allSubscriptions(self):
         """
         Yields tuples ``(required, provided, value)`` for all the
--- a/src/zope/interface/adapter.py
+++ b/src/zope/interface/adapter.py
@@ -293,11 +293,14 @@ class BaseAdapterRegistry(object):
 
         self.changed(self)
 
-    def registered(self, required, provided, name=u''):
+    def _find_leaf(self, byorder, required, provided, name):
+        # Find the leaf value, if any, in the *byorder* list
+        # for the interface sequence *required* and the interface
+        # *provided*, given the already normalized *name*.
+        #
+        # If no such leaf value exists, returns ``None``
         required = tuple([_convert_None_to_Interface(r) for r in required])
-        name = _normalize_name(name)
         order = len(required)
-        byorder = self._adapters
         if len(byorder) <= order:
             return None
 
@@ -311,6 +314,14 @@ class BaseAdapterRegistry(object):
             components = d
 
         return components.get(name)
+
+    def registered(self, required, provided, name=u''):
+        return self._find_leaf(
+            self._adapters,
+            required,
+            provided,
+            _normalize_name(name)
+        )
 
     @classmethod
     def _allKeys(cls, components, i, parent_k=()):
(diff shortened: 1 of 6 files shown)


src/zope/interface/adapter.py
```diff
@@ -430,6 +430,26 @@ def subscribe(self, required, provided, value):
 
         self.changed(self)
 
+    def subscribed(self, required, provided, subscriber):
```
jamadden, line 433: This method is almost identical to `registered`, except for which list it operates on (`_adapters` or `_subscribers`), the hardcoded name, and the last line. 

Now that I see that, it's probably worth refactoring to a single implementation.

jamadden:
Hi @mgedmin ! You 👍 the original issue (#230). Is this what you were looking for?

src/zope/interface/interfaces.py
```diff
@@ -997,19 +997,62 @@ def subscribe(required, provided, subscriber): # pylint:disable=arguments-differ
            Subscribers have no names.
         """
 
+    def subscribed(required, provided, subscriber):
+        """
+        Check whether the object *subscriber* is registered directly
+        with this object via a previous
+        call to ``subscribe(required, provided, subscriber)``.
```
mgedmin, line 1004: Word-wrapping feels a little bit weird here.  Any particular reason?

src/zope/interface/registry.py
```diff
@@ -505,6 +505,72 @@ def unregisterHandler(self, factory=None, required=None, name=u''):
...
+        By default, this method only checks for corruption. To make it
+        actually rebuild the registry, pass `True` for *rebuild*.
+
+        :param bool rebuild: If set to `True` (not the default),
+           this method will actually register and subscribe utilities
+           in the registry as needed to synchronize with the local cache.
+
+        :return: A dictionary that's meant as diagnostic data. The keys
+           and values may change over time. When called with a false *rebuild*,
+           the keys ``"needed_registered"`` and ``"needed_subscribed`` will be
```
mgedmin, line 526:
```suggestion
           the keys ``"needed_registered"`` and ``"needed_subscribed"`` will be
```

mgedmin (commented):
I cannot say that I understand what is happening here, but I see nothing obviously wrong with it.

src/zope/interface/interfaces.py
jamadden: The double-backtick block wraps as a single entity.

src/zope/interface/interfaces.py
```diff
@@ -997,19 +997,62 @@ def subscribe(required, provided, subscriber): # pylint:disable=arguments-differ
...
+
+        If the *subscriber*, or one equal to it, has been subscribed,
+        for the given *required* sequence and *provided* interface,
+        return that object. (This does not guarantee whether the *subscriber*
+        itself is returned, or an object equal to it.)
+
+        If it has not, return ``None``.
```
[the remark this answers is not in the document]
jamadden: The new `subscribed()` call is consistent with the existing `registered()` call.

src/zope/interface/registry.py
jamadden: Thanks!

src/zope/interface/interfaces.py
mgedmin: Yes, but `call to` is outside the double-ticked block and would easily fit on the previous line.

jamadden: I just realized that's what you meant and was committing a change as you commented :)

jamadden:
> I see nothing obviously wrong with it.

I think I'll take that as approval and merge this in the next hour or so unless there's more feedback 😄 

Thank you for the review.

PR #232, opened 2021-03-15
Make ZOPE_INTERFACE_STRICT_IRO imply ZOPE_INTERFACE_LOG_CHANGED_IRO.

### LINKED ISSUE #229: `directlyProvides` doesn't log changed resolution orders
When debugging https://github.com/zopefoundation/zc.resourcelibrary/issues/6, we discovered a resolution order that changed between the legacy and C3 orderings. This resolution order was created by `directlyProvides()`, not statically. For whatever reason, `directlyProvides` doesn't respect the environment variables that help debug changed resolution orders (`ZOPE_INTERFACE_LOG_CHANGED_IRO` and its superset `ZOPE_INTERFACE_USE_LEGACY_IRO`), so this had to be tracked down manually.

Can we make `directlyProvides` cheaply (!) respect those?

jamadden:
This helps diagnosing problems when just strict mode isn't enough, and strict mode is the one that's being used most of the time.

Fixes #229

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,14 @@
 5.3.0 (unreleased)
 ==================
 
+- Make the debug environment variable ``ZOPE_INTERFACE_STRICT_IRO``
+  imply ``ZOPE_INTERFACE_LOG_CHANGED_IRO``. See
+  `issue 229 <https://github.com/zopefoundation/zope.interface/issues/229>`_.
+
+- Improve the repr of ``zope.interface.Provides`` to remove ambiguity
+  about what is being provided. This is especially helpful diagnosing
+  IRO issues.
+
 - Allow subclasses of ``BaseAdapterRegistry`` (including
   ``AdapterRegistry`` and ``VerifyingAdapterRegistry``) to have
   control over the data structures. This allows persistent
--- a/src/zope/interface/declarations.py
+++ b/src/zope/interface/declarations.py
@@ -750,10 +750,11 @@ class Provides(Declaration):  # Really named ProvidesClass
         Declaration.__init__(self, *(interfaces + (implementedBy(cls), )))
 
     def __repr__(self):
-        return "<%s.%s for %s>" % (
+        return "<%s.%s for instances of %s providing %s>" % (
             self.__class__.__module__,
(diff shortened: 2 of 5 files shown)


jamadden:
I imagine this would have helped diagnosing https://github.com/zopefoundation/zc.resourcelibrary/issues/6 sooner. What do you think, @mgedmin? 

My primary concern is that this may log something to the zope.interface.ro logger, and if doctests are both using ZOPE_INTERFACE_STRICT_IRO *and* have an indiscriminate log handler installed for testing (e.g., `zope.testing.loggingsupport.Handler/InstalledHandler`) they might get unexpected output and break. That doesn't seem very likely to me, but with doctests, 🤷

jamadden:
Having slept (a little) on it, I'm not sure this is the best way to go, as it doesn't really add any new functionality and muddies some existing functionality.

Other options include:

1. Update the documentation to recommend setting `LOG_CHANGED` if `STRICT_IRO` doesn't diagnose your issue.
2. Add a new option (or a new value for `STRICT_IRO`) that forbids the IRO from being different between legacy and C3, instead raising an error, like `STRICT_IRO` does if it can't use C3. 

(1) is easy, but may not be helpful. (See next point for why.)

(2) is also pretty easy, but may have many benign positives that make it hard to use until it spreads throughout the ecosystem. For example, right now, there's logging from BTrees about a changed order for `ISet`, but AFAIK, it really makes no difference. There's even logging from this package about changed IROs for `IFullMapping` and `list`!

That concern could be mitigated if the option allowed a filter of packages to produce errors for. We could set that in each package's test environment: `ZOPE_INTERFACE_STRICT_CHANGED_IRO=BTrees` to be sure it doesn't happen. But again, I'm not sure how much value that adds over just asking for the logs when there's a problem. The long term goal is to make strict C3 mode the default, and then the legacy order is irrelevant. 

I'm considering closing this PR in favor of one that just does (1).

jamadden:
Closed in favor of #234

icemac:
@jamadden Can we delete the branch or is it still needed?

jamadden:
It's gone now. Thank you.

PR #233, opened 2021-03-17
Update the Adaptation docs to be more concrete.

jamadden:
This should help provide better motivating use cases compared to the what the docs called the "silly" example of converting one number to another. 

Examples inspired by https://glyph.twistedmatrix.com/2021/03/interfaces-and-protocols.html

Also some minor typo fixes and updates to comments. (One of which is for https://github.com/zopefoundation/zope.component/pull/53.)

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -861,7 +861,17 @@ And the list will be filled with the individual exceptions:
 Adaptation
 ==========
 
-Interfaces can be called to perform adaptation.
+Interfaces can be called to perform *adaptation*. Adaptation is the
+process of converting an object to an object implementing the
+interface. For example, in mathematics, to represent a point in space
+or on a graph there's the familiar Cartesian coordinate system using
+``CartesianPoint(x, y)``, and there's also the Polar coordinate system
+using ``PolarPoint(r, theta)``, plus several others (homogeneous,
+log-polar, etc). Polar points are most convenient for some types of
+operations, but cartesian points may make more intuitive sense to most
+people. Before printing an arbitrary point, we might want to *adapt* it
+to ``ICartesianPoint``, or before performing some mathematical
+operation you might want to adapt the arbitrary point to ``IPolarPoint``.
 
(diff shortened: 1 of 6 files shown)


jensens:
I often explain adaption with the [abstract factory design pattern](https://en.wikipedia.org/wiki/Abstract_factory_pattern). At least for those who learned about design patterns this simplifies understanding a lot.

jamadden:
Agreed, that can be a good way to explain it. 

I'm not sure that's the appropriate approach for this particular piece documentation though, because it's, well, abstract 😄 

The original was trying to work through some concrete examples, with the intent of showing how things work and how you can use them. But upon re-reading it as a result of asking glyph to link to it, I found that the "concrete" examples, well, weren't (converting 0 into 42? Why would I want to do that?). I wanted to improve that and make the material more engaging so that hopefully people can more easily imagine how they might use it in their projects. Using points was the first example that came to mind (and it seemed pretty good because there are many representations for the same thing, but they all work slightly differently), but if there's a better one I'd be happy to make adjustments.

docs/README.rst
```diff
@@ -870,52 +880,62 @@ If an object cannot be adapted, then a ``TypeError`` is raised:
...
 .. doctest::
 
-  >>> @zope.interface.implementer(I)
-  ... class C(object):
-  ...     pass
+  >>> @zope.interface.implementer(ICartesianPoint)
+  ... class CartesianPoint(object):
+  ...     """The default cartesian point is the origin."""
+  ...     def __init__(self, x=0, y=0):
+  ...         self.x = x; self.y = y
```
icemac, line 909:
```suggestion
  ...         self.x = x
  ...         self.y = y
```

icemac (approved):
LGTM.

docs/README.rst
jamadden: Hehe. You caught me 😄

jamadden:
Thank you for the reviews and comments.

PR #234, opened 2021-03-17
Update documentation and Provides repr for better debugging.

### LINKED ISSUE #229: `directlyProvides` doesn't log changed resolution orders

jamadden:
Fixes #229.

Replaces #232

### DIFF
--- a/src/zope/interface/ro.py
+++ b/src/zope/interface/ro.py
@@ -45,6 +45,10 @@ ZOPE_INTERFACE_STRICT_IRO
     If this is set to "1", any attempt to use :func:`ro` that would produce a non-C3
     ordering will fail by raising :class:`InconsistentResolutionOrderError`.
 
+.. important::
+
+   ``ZOPE_INTERFACE_STRICT_IRO`` is intended to become the default in the future.
+
 There are two environment variables that are independent.
 
 ZOPE_INTERFACE_LOG_CHANGED_IRO
@@ -56,6 +60,25 @@ ZOPE_INTERFACE_USE_LEGACY_IRO
     legacy IRO will be used instead. This is a temporary measure and will be removed in the
     future. It is intended to help during the transition.
     It implies ``ZOPE_INTERFACE_LOG_CHANGED_IRO``.
+
+.. rubric:: Debugging Behaviour Changes in zope.interface 5
+
+Most behaviour changes from zope.interface 4 to 5 are related to
+inconsistent resolution orders. ``ZOPE_INTERFACE_STRICT_IRO`` is the
+most effective tool to find such inconsistent resolution orders, and
+we recommend running your code with this variable set if at all
+possible. Doing so will ensure that all interface resolution orders
+are consistent, and if they're not, will immediately point the way to
+where this is violated.
+
+Occasionally, however, this may not be enough. This is because in some
+cases, a C3 ordering can be found (the resolution order is fully
+consistent) that is substantially different from the ad-hoc legacy
+ordering. In such cases, you may find that you get an unexpected value
+returned when adapting one or more objects to an interface. To debug
+this, *also* enable ``ZOPE_INTERFACE_LOG_CHANGED_IRO`` and examine the
+output. The main thing to look for is changes in the relative
+positions of interfaces for which there are registered adapters.
 """
 from __future__ import print_function
 __docformat__ = 'restructuredtext'
--- a/src/zope/interface/tests/test_declarations.py
+++ b/src/zope/interface/tests/test_declarations.py
@@ -1305,10 +1305,16 @@ class ProvidesClassTests(unittest.TestCase):
         self.assertRaises(AttributeError, _test)
 
     def test__repr__(self):
-        inst = self._makeOne(type(self))
+        from zope.interface.interface import InterfaceClass
+        IFoo = InterfaceClass("IFoo")
+
+        inst = self._makeOne(type(self), IFoo)
         self.assertEqual(
             repr(inst),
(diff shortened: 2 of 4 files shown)


src/zope/interface/tests/test_declarations.py
```diff
@@ -1305,13 +1305,18 @@ def _test():
...
     def test__repr__(self):
-        inst = self._makeOne(type(self))
+        from zope.interface.interface import InterfaceClass
+        IFoo = InterfaceClass("IFoo")
+
+        inst = self._makeOne(type(self), IFoo)
         self.assertEqual(
             repr(inst),
-            "<zope.interface.Provides for %r>"  % type(self)
+            "<zope.interface.Provides for instances of %r providing %s>"  % (
```
icemac, line 1314: I'd prefer to see the complete text here. Is this possible?
Currently the test seems to be is nearly a repetition of the code in the method under test.

src/zope/interface/ro.py
```diff
@@ -56,6 +60,25 @@
...
+where this is violated.
+
+Occasionally, however, this may not be enough. This is because in some
+cases, a C3 ordering can be found (the resolution order is fully
+consistent) that is substantially different from the ad-hoc legacy
+ordering. In such cases, you may find that you get an unexpected value
+returned when adapting one or more objects to an interface. To debug
+this, *also* enable ``ZOPE_INTERFACE_LOG_CHANGED_IRO`` and examine the
+output. The main thing to look for is changes in the relative
+positions of interfaces for which there are registered adapters.
```
icemac, line 81: Thank you for clearly documenting this.

icemac (approved):
Mostly LGTM.

src/zope/interface/tests/test_declarations.py
jamadden: Great point. I'll make a change.

jamadden: Given `spec = Provides(AClass, IFoo, IBar)`, the repr will be something like 
```
<zope.interface.Provides for instances of <class 'some.module.AClass'> providing (<InterfaceClass some.module.interfaces.IFoo>, <InterfaceClass some.module.interfaces.IBar>)>
```

Informative, but not terse.

Basically none of these objects have a `__str__` method.

`__repr__` is meant to "look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment)."

In contrast `__str__`  computes "the 'informal' string representation of an object. This differs from `__repr__()` in that it does not have to be a valid Python expression: a more convenient or concise representation may be used instead. The return value must be a string object."

For `Provides`, it would be pretty easy to give a `repr` that actually can be evaluated (in many cases) outputting something like
```
Provides(AClass, IFoo, IBar)
```

Similarly, interfaces could have a `__str__` that just outputs `__module__.__name__`. 

That would transform reports from 
```
Object <InterfaceClass BTrees.Interfaces.ISet> has different legacy and C3 MROs:
  Legacy RO (len=7)                                  C3 RO (len=7; inconsistent=no)
  ====================================================================================================
    <InterfaceClass BTrees.Interfaces.ISet>            <InterfaceClass BTrees.Interfaces.ISet>
    <InterfaceClass BTrees.Interfaces.IKeySequence>    <InterfaceClass BTrees.Interfaces.IKeySequence>
  - <InterfaceClass BTrees.Interfaces.ISized>
    <InterfaceClass BTrees.Interfaces.ISetMutable>     <InterfaceClass BTrees.Interfaces.ISetMutable>
    <InterfaceClass BTrees.Interfaces.IKeyed>          <InterfaceClass BTrees.Interfaces.IKeyed>
    <InterfaceClass BTrees.Interfaces.ICollection>     <InterfaceClass BTrees.Interfaces.ICollection>
                                                     + <InterfaceClass BTrees.Interfaces.ISized>
    <InterfaceClass zope.interface.Interface>          <InterfaceClass zope.interface.Interface>
```

to
```
Object <InterfaceClass BTrees.Interfaces.ISet> has different legacy and C3 MROs:
  Legacy RO (len=7)                                  C3 RO (len=7; inconsistent=no)
  ==================================================================================
    BTrees.Interfaces.ISet            BTrees.Interfaces.ISet
    BTrees.Interfaces.IKeySequence    BTrees.Interfaces.IKeySequence
  - BTrees.Interfaces.ISized
    BTrees.Interfaces.ISetMutable     BTrees.Interfaces.ISetMutable
    BTrees.Interfaces.IKeyed          BTrees.Interfaces.IKeyed
    BTrees.Interfaces.ICollection     BTrees.Interfaces.ICollection
                                    + BTrees.Interfaces.ISized
    zope.interface.Interface          zope.interface.Interface
```

(With similar benefits when `Provides` is present.)

Since the changes in `Provides` here mean that there *might* have to be doctest updates, and since adding a `__str__` mostly shouldn't (but might!) cause doctest updates, this might be a good time to make such a change.

What do you think?

jamadden: Moved discussion to issue #236.

jamadden:
Thanks for the review!

PR #238, opened 2021-04-01
Make Declaration.__add__ try harder to produce consistent resolution orders

### LINKED ISSUE #193: Declaration.__add__ easily produces inconsistent RO
Noticed while working on https://github.com/zopefoundation/zope.schema/issues/80.

It simply takes itself as the `__bases__`,  and whatever is unique to its second argument it tacked on, ignoring their relative order. If the two arguments partially overlap, this is almost guaranteed to produce an inconsistent RO.

jamadden:
By moving things from the RHS to the front of the list if they already extend something from the LHS.

Fixes #193

The [changes in the documentation](https://github.com/zopefoundation/zope.interface/compare/issue193?expand=1#diff-8ef5cd6985c269c8fa159314befb6fc8082c919e8f75727958b7e2fa36e2294a) are a good example of the kind of weird things that used to happen but which should now make more sense. Previously, you could get `IRoot2` ahead of `IDerived2`, which is inconsistent.

I'm not 100% positive this covers all possible cases (deeply nested hierarchies, branching at weird places) but it does fix the easy/common ones that I've seen in the real world. It can be improved as needed given further examples.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,14 @@
   shown as ``classImplements(list, IMutableSequence, IIterable)``. See
   `issue 236 <https://github.com/zopefoundation/zope.interface/issues/236>`_.
 
+- Make ``Declaration.__add__`` (as in ``implementedBy(Cls) +
+  ISomething``) try harder to preserve a consistent resolution order
+  when the two arguments share overlapping pieces of the interface
+  inheritance hierarchy. Previously, the right hand side was always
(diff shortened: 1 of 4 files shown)


jamadden:
GitHub actions was having some issues with its Mac runners this morning, but they seem to be resolved and this PR is now green.

icemac (approved):
LGTM from reading the code.

jamadden:
Thank you!

Issue #246, opened 2021-08-30
Unable to install zope-interface with python 3.9 for v4.7.1 zope-interface


rlam3:
### What I did:

Tried installing humming bot from source and it looks like it doesn't support python 3.9
https://github.com/CoinAlpha/hummingbot-docs/blob/9dbe2cab183de55cfca7c9316a11abcb8b21f89f/content/installation/mac.mdx

### What actually happened:

```
Collecting zope-interface==4.7.1
  Using cached zope.interface-4.7.1.tar.gz (151 kB)

Pip subprocess error:
  Running command git clone -q https://github.com/CoinAlpha/python-signalr-client.git /private/var/folders/th/c91sj7ss2m173ylmh8xf7g_m0000gn/T/pip-req-build-9hq7zeh3
    ERROR: Command errored out with exit status 1:
     command: /Users/XXXXXX/opt/anaconda3/envs/hummingbot/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/th/c91sj7ss2m173ylmh8xf7g_m0000gn/T/pip-install-g3wm0rm6/zope-interface_ddc133df7b2141cc81aee1787b3f7657/setup.py'"'"'; __file__='"'"'/private/var/folders/th/c91sj7ss2m173ylmh8xf7g_m0000gn/T/pip-install-g3wm0rm6/zope-interface_ddc133df7b2141cc81aee1787b3f7657/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/th/c91sj7ss2m173ylmh8xf7g_m0000gn/T/pip-pip-egg-info-lrgs6tkz
         cwd: /private/var/folders/th/c91sj7ss2m173ylmh8xf7g_m0000gn/T/pip-install-g3wm0rm6/zope-interface_ddc133df7b2141cc81aee1787b3f7657/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
[...]

PR #247, opened 2021-08-30
Update setup.py

rlam3:
Using python 3.9 breaks

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,6 +144,8 @@ setup(name='zope.interface',
           '!=3.2.*',
           '!=3.3.*',
           '!=3.4.*',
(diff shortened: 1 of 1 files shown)


jugmac00:
Hi @rlam3,
this PR makes no sense, see #246.

jensens:
I was about to add a similar comment.

icemac:
@rlam3 As you can see from the latest GitHub Actions run, Python 3.9 is supported, it can be installed and is successfully tested. There is no need for such drastic measures you proposed in this PR.

See https://github.com/zopefoundation/zope.interface/actions/runs/1179243067

PR #248, opened 2021-09-02
Add some more entries to zope.interface.interfaces.__all__

cjwatson:
`ComponentLookupError`, `Invalid`, `Registered`, and `Unregistered` are
all documented
(https://zopeinterface.readthedocs.io/en/latest/api/components.html), so
it seems odd not to list them in `__all__`.  `ObjectEvent` isn't
documented there, but it seems reasonable to export it as a base class
for similar events.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@
 
 - Add missing Trove classifier showing support for Python 3.9.
 
+- Add some more entries to ``zope.interface.interfaces.__all__``.
+
 
 5.4.0 (2021-04-15)
 ==================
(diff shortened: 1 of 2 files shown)


icemac:
What is your goal you want to achieve by this PR?

cjwatson:
In Launchpad we use an import hook that warns us if we import from a module that defines `__all__` and the name we're importing isn't in `__all__`, to encourage us to use only documented facilities of modules.  I'd like to be able to use these names (since after all they're described in the API documentation) without having to add manual exceptions to that hook.

icemac (approved):
LGTM. Thank you for the explanation of the rationale.

PR #250, opened 2021-11-19
Add support for Python 3.10

### LINKED ISSUE #245: .meta.toml or the config template is out of sync with actual files
I ran the config-package.py from zopefoundation/meta and discovered that it makes unexpected changes:
- the entire `docs:` section gets removed from .github/workflows/tests.yml
- a `setenv = PURE_PYTHON=1` shows up in one of the tox testenv sections
- flake8 gets enabled unexpectedly, fixed in https://github.com/zopefoundation/meta/pull/111

and then the tool runs tox and half of the environments fail with mysterious `ModuleNotFoundError: No module named 'zope.testing'` errors.

This should not be happening.  It should be safe to re-generate the CI files from the template at any time without breaking anything.

I don't currently have the time/energy to look into this.

icemac:
Additionally fixes #245.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,24 +92,26 @@ jobs:
     # with `test`, and `docs` must use a subset.
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         python-version:
-          - 2.7
-          - 3.5
-          - pypy-2.7
-          - pypy-3.6
-          - 3.6
-          - 3.7
-          - 3.8
-          - 3.9
+          - "2.7"
+          - "3.5"
+          - "pypy-2.7"
+          - "pypy-3.7"
+          - "3.6"
+          - "3.7"
+          - "3.8"
+          - "3.9"
+          - "3.10"
         os: [ubuntu-20.04, macos-latest]
         exclude:
           - os: macos-latest
-            python-version: pypy-2.7
+            python-version: "pypy-2.7"
           - os: macos-latest
-            python-version: pypy-3.6
+            python-version: "pypy-3.7"
           - os: macos-latest
-            python-version: 3.5
+            python-version: "3.5"
 
     steps:
       - name: checkout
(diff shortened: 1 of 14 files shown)


jensens (approved):
LGTM

.github/workflows/tests.yml
```diff
@@ -415,13 +416,13 @@ jobs:
         # The 2010 image is the most recent spec that comes with Python 2.7,
         # and only up through the tag 2021-02-06-3d322a5
         env:
-          DOCKER_IMAGE: quay.io/pypa/${{ matrix.image }}:2021-02-06-3d322a5
+          DOCKER_IMAGE: quay.io/pypa/${{ matrix.image }}
```
jamadden, line 419: This change appears to be dropping support for building Python 2.7 wheels. Is that deliberate?

src/zope/interface/common/tests/test_collections.py
```diff
@@ -93,6 +94,10 @@ def test_non_iterable_UserDict(self):
         # It's imported because...? Coverage imports it, but why do we have it without
         # coverage?
         'Row',
+        # In Python 3.10 ``array.array`` appears as ``IMutableSequence`` but it
+        # does not provide a ``clear()`` method and it cannot be instantiated
+        # using ``array.array()``.
+        array.array,
```
jamadden, line 100: I wondered if here and below it would be good to only add these entries in the described cases (i.e., only on Python 3.10), but it's probably not worth the effort, because it probably just depends on what gets imported when, and that could change in a minor release.

jamadden (commented):
Except for the array failure on Windows (what is going on there I wonder?) and dropping building 2.7 wheels, this seems fine to me.

.github/workflows/tests.yml
icemac: That old pinned version does not contain a Python 3.10. The pin was removed in https://github.com/zopefoundation/meta/commit/3e329394c636b6ef844221acb7f3ba1a0b496e3c. I do not think it is worth the effort to have multiple docker file versions to get wheels for outdated Python versions. (Python 3.5 are also no longer built.)

src/zope/interface/common/tests/test_collections.py
icemac: Yes, we would guard against Python 3.10+ but older versions do not have `array.array` in the list either.

PR #265, opened 2022-12-10
Fix test deprecation warning on Python 3.11

cjwatson:
Python 3.11 warns:

  DeprecationWarning: It is deprecated to return a value that is not
  None from a test case (<bound method Test_c3_ro.test_complex_diamond
  of <zope.interface.tests.test_ro.Test_c3_ro
  testMethod=test_complex_diamond>>)

Rearrange slightly to avoid this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 5.5.3 (unreleased)
 ==================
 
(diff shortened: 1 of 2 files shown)


mgedmin (approved):
:+1:

PR #274, opened 2023-08-24
improve `declarations.Provides` documentation

d-maurer:
The signature of `declarations.Provides` is given as `*interfaces`. This is wrong; it actually is `cls, *interfaces` where *cls* is the class of the instance for which `Provides` should provide a declaration. This parameter is necessary to avoid the construction of inconsistent resolution orders.

The PR does not change the code but improves the `Provides` docstring and especially documents the correct signature there.


jugmac00:
As you mention that the signature is wrong, what keeps you from updating it?

Issue #279, opened 2023-11-01
heads-up: test failures with Python 3.13~alpha1 - indentation doesn't match the expected


befeleme:
### What I did:

In Fedora Linux, we're attempting to rebuild the system with the early Python 3.13 releases.
The test failure happens with Python 3.13~alpha1 during the RPM build.

In [What's new in Python 3.13](https://docs.python.org/dev/whatsnew/3.13.html#removed) there's:

> Compiler now strip indents from docstrings. This will reduce the size of [bytecode cache](https://docs.python.org/dev/glossary.html#term-bytecode) (e.g. .pyc file). For example, cache file size for sqlalchemy.orm.session in SQLAlchemy 2.0 is reduced by about 5%. This change will affect tools using docstrings, like [doctest](https://docs.python.org/dev/library/doctest.html#module-doctest). (Contributed by Inada Naoki in [gh-81283](https://github.com/python/cpython/issues/81283).)

### What I expect to happen:
Successful build

### What actually happened:
```
+ python3 -m unittest discover -s zope/interface -t .
................................s......................s.........................................................ss.....s.............s.s..................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................F..........F....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
======================================================================
FAIL: test_asReStructuredText_empty_with_multiline_docstring (zope.interface.tests.test_document.Test_asReStructuredText.test_asReStructuredText_empty_with_multiline_docstring)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILDROOT/python-zope-interface-6.0-2.fc40.x86_64/usr/lib64/python3.13/site-packages/zope/interface/tests/test_document.py", line 298, in test_asReStructuredText_empty_with_multiline_docstring
    self.assertEqual(self._callFUT(IEmpty), EXPECTED)
AssertionError: '``IE[39 chars]n \n It can be used to annotate any class or o[66 chars]\n\n' != '``IE[39 chars]n \n             It can be used to annotate an[90 chars]\n\n'
  ``IEmpty``
  
   This is an empty interface.
   
-  It can be used to annotate any class or object, because it promises
+              It can be used to annotate any class or object, because it promises
? ++++++++++++
-  nothing.
+              nothing.
  
   Attributes:
  
   Methods:
======================================================================
[...]

PR #281, opened 2023-12-18
Make test_document compatible with Python 3.13

frenzymadness:
In Python 3.13, compiler strips indents from docstrings. See https://github.com/python/cpython/issues/81283

Fixes: https://github.com/zopefoundation/zope.interface/issues/279


icemac:
Thank you for this PR. I am going to merge it even though we do not yet run tests on Python 3.13 in GHA.

icemac (approved):

PR #267, opened 2023-01-09
Add support for ``typing.Union``'s syntax sugar (PEP 604)

gjo:
After PEP 604, we can use `int | str | None` on typing annotations instead of `typing.Optional[typing.Union[int, str]]`, however `zope.interface.Interface` doesn't support this feature.

```
>>> from zope.interface import Interface
>>> class I(Interface):
...     pass
...
>>> I | None
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'InterfaceClass' and 'NoneType'
>>>
```

This patch add support for `typing.Union`'s syntax sugar to `zope.inteface.Interface`.

PR #285, opened 2024-02-15
Add support to use the pipe (`|`) syntax for `typing.Union`.

### LINKED ISSUE #280: Consider adding support for `__or__`/`__ror__` to `IntefaceClass`
##  FEATURE REQUEST

Currently interfaces created with zope.interface do not support the Python3.10 shorthand notation for union types, it makes sense that this is not supported by default, since subclassing `Interface`  creates instances of `InterfaceClass` which is not a subclass of `type`.

### What I did:

```python
class IFoo(Interface):
    pass

x: IFoo | None = None
```

### What I expect to happen:

`IFoo | None` results in either `typing.Union[IFoo, None]` (or `types.UnionType(IFoo, None)` if it is implemented in a C extension)

### What actually happened:

```
TypeError: unsupported operand type(s) for |: 'InterfaceClass' and 'NoneType'
```

### What version of Python and Zope/Addons I am using:

Python 3.10
zope.interface 6.1

The naive fix for this should be straightforward, I don't think we need to worry about emitting type errors for invalid union members. The standard library is in charge of `Union` and making it work correctly, so we don't need to do anything special.

```python
from typing import Union

...

class InterfaceClass(_InterfaceClassBase):

    ...

    def __or__(self, other):
        return Union[self, other]

    def __ror__(self, other):
        return Union[other, self]

    ...
```
This work by David Salvisberg is marked with [CC0 1.0 Universal](http://creativecommons.org/publicdomain/zero/1.0?ref=chooser-v1).

icemac:
Fixes #280.
Closes #267.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@
 
 - Add preliminary support for Python 3.13 as of 3.13a3.
 
+- Add support to use the pipe (``|``) syntax for ``typing.Union``.
+  (`#280 <https://github.com/zopefoundation/zope.interface/issues/280>`_)
+
+
 6.1 (2023-10-05)
 ================
 
--- a/src/zope/interface/interface.py
+++ b/src/zope/interface/interface.py
@@ -18,6 +18,7 @@ import sys
 from types import MethodType
 from types import FunctionType
 import weakref
+from typing import Union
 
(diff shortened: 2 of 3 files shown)


dataflake (approved):

icemac:

PR #296, opened 2024-05-18
Update meta/config to publish Windows wheels

dataflake:
Update to an (unreleased) meta/config template that adds a publishing step for Windows wheels when git tags are set.


dataflake:
Thanks, go ahead and do that when you have a moment.

PR #297, opened 2024-05-21
refactor: toward multi-phase module init (WIP)

tseaver:
Per PEP 489.

With the aim of supporting sub-interpreters, and eventually free-threading.

Some segfaults remain, and the C module needs another pass through updating names, docstrings, comments etc.

### DIFF
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -115,12 +115,17 @@ SpecBase_clear(SpecBase* self)
     return 0;
 }
 
+#undef LOG
+#define LOG(msg)
+//#define LOG(msg) printf((msg))
 static void
 SpecBase_dealloc(SpecBase* self)
 {
+    PyTypeObject *tp = Py_TYPE(self);
     PyObject_GC_UnTrack((PyObject *)self);
     SpecBase_clear(self);
-    Py_TYPE(self)->tp_free(OBJECT(self));
+    tp->tp_free(OBJECT(self));
+    Py_DECREF(tp);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


tseaver:
*Moved here from the Python Discourse server because too many links*.

I'm trying to follow along with the guide linked by @ngoldbaum, as well as the `xxlimited` example reference by @erlendaasland while trying to port the [`_zope_interface_c_optimizations` extension module](https://github.com/zopefoundation/zope.interface/blob/master/src/zope/interface/_zope_interface_coptimizations.c) to multi-phase initialization and heap types.

One thing I don't see covered in either the guide or the example module is how to deal with weakrefs in the New World Order.  The existing `_z_i_c` module's `Spec` [holds an explicit `weakreflist`](https://github.com/zopefoundation/zope.interface/blob/e8ad972f7b942c0bec6155bcb62ccfc7cef4bf11/src/zope/interface/_zope_interface_coptimizations.c#L319) and [clears it in `Spec_dealloc`](https://github.com/zopefoundation/zope.interface/blob/)e8ad972f7b942c0bec6155bcb62ccfc7cef4bf11/src/zope/interface/_zope_interface_coptimizations.c#L373C1-L375C6).  Of course, it [sets the `tp_weaklistoffset` field in its type definition](https://github.com/zopefoundation/zope.interface/blob/e8ad972f7b942c0bec6155bcb62ccfc7cef4bf11/src/zope/interface/_zope_interface_coptimizations.c#L508).

The [docs for `tp_members`](https://docs.python.org/3/c-api/structures.html#accessing-attributes-of-extension-types) discourage the use of `tp_weaklistoffset` in favor of `Py_TPFLAGS_MANAGED_WEAKREF`, but I cannot find any description of the difference in the semantics.  Given that my updated module can produce segfaults where the old version did not, and that the other refcount stuff is identical between them, I suspect that I'm not handling the weakrefs case correcty.

tseaver:
Superseded by #301.

PR #300, opened 2024-05-23
Attempt to fix aarch64 by only testing standard interpreters

dataflake:
Application of https://github.com/zopefoundation/meta/pull/242


icemac (approved):
It works, so let's merge it.

PR #301, opened 2024-05-23
Rework PEP489: multi-phase module initialization, heap based types

tseaver:
This PR is a reworking of PR #297, with a better final outcome:  the test suite passes on Python 3.11, 3.12, and 3.13!

Unfortunately, the dance required to juggle module state between dynamically-allocated modules and heap-allocated extension types requires APIs which are too new for all our supported Python versions.

Assuming we have no big plans for the C module, this branch could likely sit around for a bit.  Or we could choose to merge it as part of a major version (`7.0`?), dropping support for the older versions all at once, perhaps with a "security fixes" `6.x` maintenance branch.  I haven't yet played with running it under sub-interpreters or the no-gil interpreter, but will likely do some experimentation with them later.


dataflake (commented):
I'm afraid I can't provide any meaningful input here.

icemac:
@tseaver Thank you for this PR even though it is way beyond my understanding of C code.
Does your description mean the changes are Python 3.13+ only?

tseaver:
@icemac 
> Does your description mean the changes are Python 3.13+ only?

Nope, the current changes work on 3.11+.  I've stubbed my toe a couple of times trying to work out how to solve the problems with the older versions'  C APIs, but have given up for the moment.

The motivation is laying groundwork for using `zope.interface` in configurations which the current static initialization blocks, most notably sub-interpreters (PEP 554 / 734, with some existing support going back a long while) and `--no-gil` (only available in 3.13+).  I would particularly like to experiment with running a single-process web app with multiple sub-interpreters, to see if it scales better (or is easier to manage) than the current default practice of running a cluster of webapp processes.

I should probably mark the PR as a draft.

PR #314, opened 2024-08-06
chore: prep 7.0.1 release

tseaver:


dataflake (approved):

dataflake:
You create and push the tag, which will build all binary packages. You yourself create the tarball and upload it. I always do `bin/buildout setup setup.py sdist` followed by `bin/twine upload dist/XXX.tar.gz` (I always have `twine` right in my virtual environment)

dataflake:
And yes, an svb commit to cap it all off

tseaver:
/me wonders why we don't have the GHA workflow push the `sdist` as well.

dataflake:
No idea, I didn't have the will to look into it yet and haven't been bothered too much.

tseaver:
@dataflake I was thinking about the new PyPI rules about tokens (they can't be scope to organizations, only projects, or else all projects).

tseaver:
Also, I almost uploaded a `zope_interface-7.0.1.tar.gz` a moment ago, as I forgot that we have to build with a pinned `setuptools`.

dataflake:
I have a command line expansion that sets up virtual environments with all those (stupid) pins in place so I hardly ever see that issue.

PR #322, opened 2024-09-02
Config with c code template acd39fc2

dataflake:
This update introduced a lot of changes so I have structured them into several commits, one per tool that was run over the code, and then some that introduced additional fixed. I hope that will help review it all.


tox.ini
```diff
@@ -32,6 +32,8 @@ commands =
 extras =
     test
     docs
+# Must use editable package due to the way tests are discovered and run
+usedevelop = true
```
tseaver, line 36: We need to fix whatever causes the failure -- this change makes the `PURE_PYTHON` testing hard and / or not actually testing what it is supposed to do.

tseaver (changes requested):

tox.ini
dataflake: I have removed the workaround so the error should show in the GHA test runs.

tseaver: While manually bisecting where the test failure happens (tests run fine without this change on `master`), I discoverd that running steps which make actual changes to the code (the currrent `lint` stuff) can break actual tests.  E.g., if I run `tox -e lint` against commit 94fd4cbf24fad0c98f4450bfea58b2e0365e470e, it introduces an actual `SyntaxError` into the code:

```bash
$ git checkout master
$ git pull
$ tox -e py38
...
  py38: OK (21.45=setup[15.50]+cmd[1.37,4.57] seconds)
  congratulations :) (21.74 seconds)
$ git checkout 94fd4cbf24fad0c98f4450bfea58b2e0365e470e
$ tox -e lint
...
$ tox - py38
...
======================================================================
ERROR: zope.interface (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: zope.interface
  File "/opt/Python-3.8.19/lib/python3.8/unittest/loader.py", line 470, in _find_test_path
    package = self._get_module_from_name(name)
  File "/opt/Python-3.8.19/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/home/tseaver/projects/Zope/Z3/zope.interface/.tox/py38/lib/python3.8/site-packages/zope/interface/__init__.py", line 53, in <module>
    from zope.interface.interface import Interface
  File "/home/tseaver/projects/Zope/Z3/zope.interface/.tox/py38/lib/python3.8/site-packages/zope/interface/interface.py", line 1172, in <module>
    from zope.interface.declarations import implementedBy
  File "/home/tseaver/projects/Zope/Z3/zope.interface/.tox/py38/lib/python3.8/site-packages/zope/interface/declarations.py", line 967
    isinstance(provides, Implements)  # noqa W503
    ^
SyntaxError: invalid syntax
```

Linting in ways that apply changes should be done manually, with the burden on the developer proposing those changes.  If a PR fails due to linting errors, there should be an easily documented step for the PR's author to perform to apply those fixes, and test them locally, before pushing to the PR.

dataflake: The branch config-with-c-code-template-acd39fc2 as it is locally on my machine does not apply any more changes when I run `bin/tox -elint`.

You are preaching to the choir, but I already explained that I was outvoted. I hate linting making changes myself.

dataflake:
Locally all tox test runs fail with tracebacks like this unless I apply the workaround of specifying `usedevelop = true` in the tox testenv. I do not know what this means.

```
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/__main__.py", line 18, in <module>
    main(module=None)
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/main.py", line 124, in parseArgs
    self._do_discovery(argv[2:])
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/main.py", line 244, in _do_discovery
    self.createTests(from_discovery=True, Loader=Loader)
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/main.py", line 154, in createTests
    self.test = loader.discover(self.start, self.pattern, self.top)
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/loader.py", line 349, in discover
    tests = list(self._find_tests(start_dir, pattern))
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/loader.py", line 414, in _find_tests
    yield from self._find_tests(full_path, pattern, namespace)
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/loader.py", line 414, in _find_tests
    yield from self._find_tests(full_path, pattern, namespace)
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/loader.py", line 414, in _find_tests
    yield from self._find_tests(full_path, pattern, namespace)
  [Previous line repeated 1 more time]
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/loader.py", line 405, in _find_tests
    tests, should_recurse = self._find_test_path(
  File "/Users/jens/src/python/Python-3.8.18/lib/python3.8/unittest/loader.py", line 458, in _find_test_path
    raise ImportError(
ImportError: 'test_builtins' module incorrectly imported from '/Users/jens/src/zope/zope.interface/.tox/py38/lib/python3.8/site-packages/zope/interface/common/tests'. Expected '/Users/jens/src/zope/zope.interface/src/zope/interface/common/tests'. Is this module globally installed?
py38: exit 1 (0.56 seconds) /Users/jens/src/zope/zope.interface> coverage run -p -m unittest discover -s src pid=21543
  py38: FAIL code 1 (3.09=setup[2.53]+cmd[0.56] seconds)
  evaluation failed :( (3.11 seconds)
```

tox.ini
tseaver: Can you point me to the issue?  The `--aggressive --aggressive` flag is particularly egregious.  Folks asking for such should be prepared to defend it (and do the work to fix what breaks).

dataflake: See https://github.com/zopefoundation/zope.interface/pull/322#issuecomment-2326706353 and the failed GHA run at https://github.com/zopefoundation/zope.interface/actions/runs/10684980548/job/29616689638. No idea why it only fails in that one manylinux run.

dataflake:
I have tried a few variation on the unittest invocation. If I change the current `coverage run -p -m unittest discover -s src {posargs}` to `coverage run -p -m unittest discover -s src/zope/interface {posargs}` then all tests run, but many %$#^&% doctests that rely on a specific text representation of the interface class will fail because that changes. Here's an example:

```
======================================================================
FAIL: test_w_None_no_bases_w_class (tests.test_declarations.Test_implementedByFallback)
----------------------------------------------------------------------
  File "/Users/jens/src/zope/zope.interface/src/zope/interface/tests/test_declarations.py", line 698, in test_w_None_no_bases_w_class
    self.assertEqual(spec.__name__,
AssertionError: 'tests.test_declarations.Foo' != 'zope.interface.tests.test_declarations.Foo'
- tests.test_declarations.Foo
+ zope.interface.tests.test_declarations.Foo
? +++++++++++++++
```

icemac:
@dataflake If the only failures are the names we could fix this whith a RENormalizer. Do you think this would be enough or are there more tests to fix which neede individual care (like the one you pasted)?

dataflake:
If I remember correctly there are 15 failing tests and many of them unit tests that all need to be touched individually. What I am more concerned is what the failure actually means. Is it even important that the string representations look exactly like the current unit tests want?

dataflake:
The PR #325 demonstrates the test changes as well as the unittest/coverage configuration changes required to make the tests themselves run **locally**. But as you can see from the GHA test run, it breaks the tests here. And **locally** the coverage test keeps failing because it thinks there's a coverage of just 32.42%.

This package is a complete one-off because the test configuration must ensure that zope.interface is not imported from another location. That's why it doesn't use zope.testrunner.

I don't know what to do about this other than to punt and just put `usedevelop = true` back into the tox configuration.

dataflake:
I have given it another try and compared how the GHA test runners (where the tests were green) work differently from the local `tox` invocation. The main difference is that they force-install the package with the tests extra just before running the tests. So if I change the tox configuration's `[testenv]` to drop `usedevelop = true` and then add `pip install -U -e .[test]` as a command step before invoking the tests then things work again. But then again this might just do what `usedevelop = true` does behind the scenes?

The GHA actions reveal another issue, on Python 3.13 the tests will reliably fail with a segmentation fault on all platforms. I am guessing that's a separate issue.

@tseaver what do you think?

davisagli:
Note: https://github.com/zopefoundation/zope.interface/pull/324 might help with the segfault. I am waiting for @tseaver to take a look at my latest comment there.

tox.ini
dataflake: The tests are now all green both locally and on GHA without `usedevelop`, so I'm closing this conversation.

dataflake:
The tests are all green, please re-review.

dataflake:
The tests have been green since I merged `master` into this branch, which contained the fixes from #328 and also the reassurance that some Python 3.13 issues were fixed in Python between rc2 and rc3. I have received very little feedback so far and don't think it helps anyone to wait any longer.

Issue #330, opened 2024-10-19
Please make the C extension build optional (e.g. via `PURE_PYTHON`)


mgorny:
## FEATURE REQUEST

Could you consider making the C extension build in `setup.py` optional, similarly to how the extensions are currently optional in tests? (i.e. if `PURE_PYTHON=1`, it wouldn't attempt to build them at all)

Our primary use case right now is Python 3.13 freethreading. Since the speedups extension does not support freethreading, importing it effectively reenables GIL for the whole program:

```pytb
>>> import zope.interface
<frozen importlib._bootstrap>:488: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'zope.interface._zope_interface_coptimizations', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

For the time being, we believe that the "lesser evil" would be to install install `zope.interface` without speedups, and keep GIL disabled for the remainder of the program. The way we see it, it would be done like:

```
$ PURE_PYTHON=1 pip install zope.interface
```

d-maurer:
Michał Górny wrote at 2024-10-18 22:55 -0700:
>## FEATURE REQUEST
>
>Could you consider making the C extension build in `setup.py` optional, similarly to how the extensions are currently optional in tests? (i.e. if `PURE_PYTHON=1`, it wouldn't attempt to build them at all)
> ...
>For the time being, we believe that the "lesser evil" would be to install install `zope.interface` without speedups, and keep GIL disabled for the remainder of the program. The way we see it, it would be done like:
>
>```
>$ PURE_PYTHON=1 pip install zope.interface
>```

Why are you interested to suppress the installation of
the C extension?
Setting the envvar `PURE_PYTHON` to a value different from `"0"`
at run (in contrast to installation) time,
should prevent the C extension import.
Have you observed that this does not work?

If it really should not work, you can delete an installed C extension.


Binary distributions of `zope.interface` contain the C extension.
I fear that (new) special logic would be necessary to suppress its
installation based on the `PURE_PYTHON` envvar.
I am against the introduction of such special logic for your quite
special use case.

mgorny:
[...]

PR #331, opened 2024-10-19
feat: Do not use C speedups on Python 3.13+ freethreading

mgorny:
Do not enable C speedups by default if running Python 3.13+ in freethreading mode (with GIL disabled).  Since the extension does not support this mode explicitly at the moment, loading it would cause CPython to reenable GIL with a warning and penalize the whole environment.

Discussed in issue #330.

------

I've signed the CLA today, but I suppose it will take some time for it to be processed.


d-maurer (commented):
Thank you for your PR!

Please add an entry in `CHANGES.rst` describing your change. It should tell that for the time being (at least until the C optimization supports freethreading) the use of the C optimization is disabled automatically in freethreading mode.

d-maurer (approved):

icemac:
@tseaver @davisagli I thought that the refactoring of the C code was done to support free threading. Am I wrong here?

davisagli:
@icemac That's my understanding too. I'll need to find some time to look more closely and see if we just need to declare now that we support it, or if more changes are also needed.

tseaver:
@davisagli 
> I'll need to find some time to look more closely and see if we just need to declare now that we support it, or if more changes are also needed.
Docs for the work needed to support free-threading:
- [Declaring FT support](https://docs.python.org/3/howto/free-threading-extensions.html#multi-phase-initialization)
- [APIs to check](https://docs.python.org/3/howto/free-threading-extensions.html#general-api-guidelines)
- [Non-official porting guide](https://py-free-threading.github.io/porting/)

PR #337, opened 2024-12-19
Update to setuptools 75.6.0

### LINKED ISSUE #333: setuptools version limit
What is the reason to limit the version of setuptools to:

https://github.com/zopefoundation/zope.interface/blob/7e0be48d15c594cc592537da2da98311017b19ab/pyproject.toml#L6

It seems to me that it builds fine also with setuptools 74.1.3

icemac:
Fixes #333

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@
 # native support. It works, but is slow.
 #
 # Another major downside: You can't just re-run the job for one part
-# of the matrix. So if there's a transient test failure that hit, say, 3.8,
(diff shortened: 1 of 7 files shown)


dataflake (approved):

jensens:
Don't we need to update `CHANGES.rst` too? Or is this handled by something else?

dataflake:
@jensens is right, the removal of Python 3.8 support requires a change log entry, I added it.

icemac:
Sorry about the forgotten changelog entry and thank you for reviewing the PR carefully.

Issue #340, opened 2025-04-02
hard-to-repro segfault in py3.8


markjm:
## BUG/PROBLEM REPORT

Hi - I am reporting a potential bug which I am having a very hard time reproducing :(. I will continue to try and find a stable reproduction but want to share what I have found so far in case someone has seen something similar before.

### What I did:

We are very rarely hitting segfaults which look like this

```
Core was generated by `/usr/bin/python /usr/local/bin/airflow db check'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fbec003037f in _zic_state_traverse (module=<optimized out>, visit=0x4f0e20, arg=0x7fbebd007810)
    at src/zope/interface/_zope_interface_coptimizations.c:2052

warning: 2052	src/zope/interface/_zope_interface_coptimizations.c: No such file or directory
(gdb) where
#0  0x00007fbec003037f in _zic_state_traverse (module=<optimized out>, visit=0x4f0e20, arg=0x7fbebd007810)
    at src/zope/interface/_zope_interface_coptimizations.c:2052
```

```
*** SIGSEGV received at time=1743611121 on cpu 25 ***
PC: @     0x7e9b7f6a037f  (unknown)  _zic_state_traverse
     @     0x7f209b8fc090  149509968  (unknown)
     @ 0x7440000000a980f6  (unknown)  (unknown)
 [2025-04-02 16:25:21,270 E 28963 28963] logging.cc:361: *** SIGSEGV received at time=1743611121 on cpu 25 ***
 [2025-04-02 16:25:21,270 E 28963 28963] logging.cc:361: PC: @     0x7e9b7f6a037f  (unknown)  _zic_state_traverse
 [2025-04-02 16:25:21,272 E 28963 28963] logging.cc:361:     @     0x7f209b8fc090  149509968  (unknown)
 [2025-04-02 16:25:21,275 E 28963 28963] logging.cc:361:     @ 0x7440000000a980f6  (unknown)  (unknown)
```

as mentioned above, i am still having trouble with an exact repro, but will post back as soon as I have found one

### What I expect to happen:

no segfault :)

I have noticed this difference in the Python Docs

python3.8
> [traverseproc](https://docs.python.org/3.8/c-api/gcsupport.html#c.traverseproc) m_traverse
[...]

PR #341, opened 2025-04-03
fix: account for potentially-null PyModule_GetState during GC operations

markjm:
Potential fix offered for #340


icemac:
Python 3.8 is no longer supported, see also me comment in #340.

PR #344, opened 2025-09-04
Switch to PEP 420 native namespace.

icemac:
- **Bumped version for breaking release.**
- **Replace ``pkg_resources`` namespace with PEP 420 native namespace.**
- **Update to native namespace.**

### DIFF
--- a/src/zope/interface/tests/test_declarations.py
+++ b/src/zope/interface/tests/test_declarations.py
@@ -15,6 +15,7 @@
 """
 import unittest
 
+from zope.interface.interface import InterfaceClass
 from zope.interface.tests import MissingSomeAttrs
 from zope.interface.tests import OptimizationTestMixin
 from zope.interface.tests import SubclassableMixin
@@ -24,6 +25,24 @@ from zope.interface.tests.test_interface import \
 
 # pylint:disable=inherit-non-class,too-many-lines,protected-access
 # pylint:disable=blacklisted-name,attribute-defined-outside-init
+
+IFoo = InterfaceClass('IFoo')
+
+
+class Foo:
+    pass
+
+
+class FooImplementedNone:
+    __implemented__ = None
+
+
+class FooNoCall:
+    __implemented__ = None
+
+    def __call__(self):
+        raise NotImplementedError()
+
 
 class _Py3ClassAdvice:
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -26,7 +26,7 @@ setenv =
     ZOPE_INTERFACE_STRICT_IRO=1
 commands =
     pip install -U -e .[test]
-    coverage run -p -m unittest discover -s src {posargs}
+    coverage run -p -m unittest discover -s src/zope {posargs}
```
icemac, line 29: Without this change no tests were discoved.
But this change changed the module names for the classes, to I had to adjust the tests.

I think we are using unittest discover here instead of `zope.testrunner` because the latter depends on `zope.interface` itself.

dataflake: I know I have seen this particular issue before, I just don't remember where and what the fix was. It feels wrong if the `__repr__` needs to be updated to something that doesn't match the actual package name anymore.

icemac: Yeah, it does not feel right, I might fix it using the import trick I used in https://github.com/zopefoundation/zope.interface/pull/344/files#diff-a91536cc6f84a12e565a527e70ed24f6cf2bec70bb095a8daf2c08698b5ad175R62-R64 – do you think this would be a better aproach?

src/zope/interface/tests/test_ro.py
```diff
@@ -298,17 +298,17 @@ def _check_handler_complex_diamond(self):
...
+  ========================================================
+    interface.tests.test_ro.A    interface.tests.test_ro.A
+    interface.tests.test_ro.B    interface.tests.test_ro.B
+  - interface.tests.test_ro.E
+    interface.tests.test_ro.C    interface.tests.test_ro.C
+    interface.tests.test_ro.D    interface.tests.test_ro.D
+                               + interface.tests.test_ro.E
+    interface.tests.test_ro.F    interface.tests.test_ro.F
+    zope.interface.Interface     zope.interface.Interface""".format(
+            name="interface.tests.test_ro.A"
```
icemac, line 311: This feels too complicated to switch back to to the previous spelling, so I left it as it is.

tox.ini
dataflake: Well, they did the job. I really don't have the depth of knowledge of zope.interface to say if they are good or bad.

PR #345, opened 2025-09-19
Make tests resilient against different ways of calling them.

icemac:
Tests fail when run in ZTK, changed them in a way so they both pass. (I had to change the way tests are called here in https://github.com/zopefoundation/zope.interface/pull/344 for PEP-420 namespaces.)

Broken test run: https://github.com/zopefoundation/zopetoolkit/actions/runs/17726051880/job/50366806813
Fixed test run with these changes:
https://github.com/zopefoundation/zopetoolkit/actions/runs/17852072197/job/50762648466

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@
 
 - Make tests resilient against different ways of calling them.
 
+- Remove run-time dependency on ``setuptools``.
+
 
 8.0 (2025-09-12)
 ================
--- a/setup.py
+++ b/setup.py
@@ -136,7 +136,6 @@ setup(
     },
     include_package_data=True,
     zip_safe=False,
-    install_requires=['setuptools'],
     python_requires='>=3.9',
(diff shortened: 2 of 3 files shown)


src/zope/interface/tests/test_ro.py
```diff
@@ -298,23 +302,38 @@ def _check_handler_complex_diamond(self):
 
         expected = """\
 Object <InterfaceClass {name}> has different legacy and C3 MROs:
-  Legacy RO (len=7)            C3 RO (len=7; inconsistent=no)
-  ========================================================
+  Legacy RO (len=7)                 C3 RO (len=7; inconsistent=no)
+  =============================================================
```
icemac, line 306:
```suggestion
  Legacy RO (len=7)            C3 RO (len=7; inconsistent=no)
  ========================================================
```

wosc (approved):

Issue #332, opened 2024-10-22
Test / build with 3.13 free-threading support


tseaver:
Docs for the work needed to support free-threading:
- [Declaring FT support](https://docs.python.org/3/howto/free-threading-extensions.html#multi-phase-initialization)
- [APIs to check](https://docs.python.org/3/howto/free-threading-extensions.html#general-api-guidelines)
- [Non-official porting guide](https://py-free-threading.github.io/porting/)

jensens:
Draft PR with a fix: #352

The C extension already has multi-phase init, per-module state, and heap types (Python >= 3.11), so the groundwork from PR #301 is solid. The remaining issues were:

1. **Missing `Py_mod_gil_not_used`** — without this slot in the module definition, Python re-enables the GIL for the entire process on import
2. **Borrowed references in cache lookups** — `PyDict_GetItem()` returns a borrowed reference; on free-threaded builds another thread could clear the cache between the return and `Py_INCREF()`, causing use-after-free. Fixed by using `PyDict_GetItemRef()` (Python 3.13+, with polyfill for older versions)
3. **Direct `ob_type` struct access** — one instance at the `_CALL_CUSTOM_ADAPT` check, replaced with `Py_TYPE()` macro

Tested locally: 1358 tests pass on both Python 3.14 and 3.14t, no GIL re-enable warning, zero performance regression on cached lookups.

PR #352, opened 2026-02-16
Fix free-threaded Python 3.14t support in C extension

jensens:
## Summary

- Declare `Py_mod_gil_not_used` in module slots so the GIL is not re-enabled on import under free-threaded Python 3.14t (PEP 703)
- Replace `PyDict_GetItem()` borrowed references with `PyDict_GetItemRef()` strong references in all cache lookup functions (`_subcache`, `_getcache`, `_lookup`, `_lookup1`, `_lookupAll`, `_subscriptions`, `implementedBy`). This prevents use-after-free when another thread clears the cache between the dict lookup return and `Py_INCREF()`.
- Use `Py_TYPE()` macro instead of direct `ob_type` struct access (incompatible with atomic type lookups in free-threaded builds)
- Add `PyDict_GetItemRef` polyfill for Python < 3.13 (compiles to the same `PyDict_GetItem` + `Py_INCREF` pattern, zero overhead)
- Add 3.14t to CI matrix and tox envlist

Tested locally on both Python 3.14 (1358 tests pass) and Python 3.14t (1358 tests pass, no GIL re-enable warning). Benchmark shows zero regression on cached adapter lookups (~109ns/call before vs ~106ns/call after).

Refs #332

### DIFF
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -30,6 +30,40 @@
 
 #define PyNative_FromString PyUnicode_FromString
 
+#if PY_VERSION_HEX < 0x030d0000
+/* PyDict_GetItemRef() was added in Python 3.13 and returns a strong
+ * (owned) reference.  This matters for free-threaded Python (PEP 703):
+ * PyDict_GetItem() returns a borrowed reference, and another thread
+ * could clear the dict between the return and the caller's Py_INCREF,
+ * leaving a dangling pointer.  PyDict_GetItemRef() performs the INCREF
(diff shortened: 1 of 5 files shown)


src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -30,6 +30,40 @@
...
+    }
+    if (PyErr_Occurred()) {
+        *result = NULL;
+        return -1;
+    }
+    *result = NULL;
+    return 0;
+}
+#define PyDict_GetItemRef _PyDict_GetItemRef
+#endif
```
tseaver: Nice!

tseaver (approved):

PR #355, opened 2026-02-17
Add CodeQL workflow with manual c-cpp build

### LINKED ISSUE #354: CodeQL c-cpp analysis failing since Nov 2025
The CodeQL c-cpp autobuild step has been failing on every run since November 16, 2025 (commit 67db09c6 - "move package metadata into `pyproject.toml`").

The autobuild tries to compile the C extensions via `setup.py`, but the setuptools version on the GitHub Actions runner doesn't understand the PEP 639 license format used in `pyproject.toml`:

```toml
license = "ZPL-2.1"
```

It expects the legacy PEP 621 table format (`{file = "..."}` or `{text = "..."}`), resulting in:

```
ValueError: invalid pyproject.toml config: `project.license`.
configuration error: `project.license` must be valid exactly by one definition (2 matches found)
```

The Python and Actions analyses are unaffected — only the c-cpp job fails.

Last successful run: https://github.com/zopefoundation/zope.interface/actions/runs/19387173039
Latest failure: https://github.com/zopefoundation/zope.interface/actions/runs/22118043874

jensens:
The default CodeQL setup's c-cpp autobuild fails because the runner's setuptools is too old for the PEP 639 license format (`license = "ZPL-2.1"`) in pyproject.toml.

Adds a custom workflow with manual build mode for c-cpp that upgrades setuptools before building. Uses codeql-action v4 (v3 is deprecated Dec 2026).

⚠️ **Important:** A repo admin must disable the default CodeQL setup in Settings > Code security *before* merging this PR, otherwise the two configurations conflict ("CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled").

Fixes #354

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -100,8 +100,8 @@ jobs:
           - os: macos-latest
             python-version: "pypy-3.11"
         include:
-          - os: ubuntu-latest
-            python-version: "3.14t"
+          - python-version: "3.14t"
+            os: ubuntu-latest
 
     steps:
       - name: checkout
@@ -259,8 +259,8 @@ jobs:
           - os: macos-latest
             python-version: "pypy-3.11"
         include:
-          - os: ubuntu-latest
-            python-version: "3.14t"
+          - python-version: "3.14t"
+            os: ubuntu-latest
 
     steps:
       - name: checkout
@@ -623,11 +623,9 @@ jobs:
           # PyPy wheels shouldn't be uploaded, remove them if present
           find dist/ -name "*pypy*" -type f -delete || true
           find dist/ -name "*none-any*" -type f -delete || true
-          # Wheels for the no-yet-supported future Python version need to go
+          # Wheels for the not-yet-supported future Python version need to go
           find dist/ -name "*3.15*" -type f -delete || true
           find dist/ -name "*cp315*" -type f -delete || true
-          # Free-threaded wheels are not published separately
-          find dist/ -name "*cp314t*" -type f -delete || true
           # For Linux, we only want the manylinux wheels
           find dist/ -name "*linux_x86_64*" -type f -delete || true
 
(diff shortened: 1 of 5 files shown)


jensens:
Btw., more information here https://docs.github.com/en/code-security/how-tos/scan-code-for-vulnerabilities/manage-your-configuration/editing-your-configuration-of-default-setup#replacing-default-setup-with-advanced-setup

dataflake:
I have disabled the default CodeQL scan in the project settings and manually re-ran the failing CodeQL tests for the PR, they're green now.

I don't have a feeling for how widespread this issue is. My knowledge about CodeQL is close to zero and I do not want to get into the business of maintaining CodeQL workflows myself. At that point I would start wondering if the added value of a working CodeQL configuration is worth the added burden of hand-maintaining those configurations.

jensens:
I do not know why CodeQL was introduced, what the goal was and what the processes behind are, like who's responsible for it. IMO the value of those external services are limited and the burden to maintain them is higher than the yield they give. That said, I just tried to fix it. We can remove it as well, but that's not my decision.

dataflake:
CodeQL was introduced by GitHub itself as some kind of "just enable this with one click and something will scan your repo for security holes". It seemed reasonable and easy at the time so it was enabled on many repos. But to this day I don't really know what it does or if it actually helps in our particular situation.

icemac:
CodeQL found some issues in the beginning when it was introduced but this was long time ago. Nowadays it usually finds nothing. It might be helpful for repos with C code but for the Python ones I see no added benefit. On the other hand I've not worked with CodeQL at at besides enabling it and ignoring false positives. Maybe now is the time to disable it again to reduce the friction it creates.

dataflake:
I'd also be in favor of just disabling it, at least wherever the standard configuration starts failing.

Even in those repositories that have C code the amount of C code is small and it hardly ever changes. It certainly does not grow.

dataflake:
OK, let's just close this PR then. Jens, I appreciate the work, but it will just be another place with a custom solution that will need maintenance and the return value of these scans is dubious.

PR #368, opened 2026-04-25
Add creating/publishing sdists

dataflake:
This was the test case for https://github.com/zopefoundation/meta/pull/413


.github/workflows/tests.yml
```diff
@@ -200,6 +200,13 @@ jobs:
           python setup.py build_ext -i
           python setup.py bdist_wheel
 
+      - name: Create zope.interface sdist
+        if: >
+          startsWith(runner.os, 'Linux')
+          && matrix.python-version == '3.14'
+        run: |
+          python setup.py sdist
```
gforcada, line 208: An alternative to not rely on `setup.py` could be using [`build`](https://pypi.org/project/build):

```python
    python -m build --sdist
```

dataflake: IMHO as long as it works there's no need to install yet another package just to create a distribution. We can revisit this when/if we get rid of setup.py entirely.

icemac (approved):
The next time I'll do a release I'll add the necessary configuration to prevent `zest.releaser` from creating an sdist. For this is good enough.

dataflake:
> The next time I'll do a release I'll add the necessary configuration to prevent `zest.releaser` from creating an sdist. For this is good enough.

Would it make sense to add a PyPI publish step to the default template? The drawback is that when you update a package with zope.meta for the first time after this change you'd have to visit the package's PyPI configuration and enabled trusted publishing.

icemac:
> Would it make sense to add a PyPI publish step to the default template?

We could switch all packages to trusted publishing, this would allow people who are not listed on PyPI as maintainers to create releases, they'll then just need the approval of one of the maintainers. I am not sure that this is a use case.

mauritsvanrees:
If you are not on PyPI as maintainer, then likely you don't have the rights to push directly to master/main either.  So you would have to both create the tag for the release, and make a PR to merge the release commits.  Then I think it is easier for everyone to let a maintainer do it all.

For Plone packages I would not want this.  (Of course they are not handled by zope.meta, so not directly relevant here.)  Biggest reason: when I make a release of a plone package, I directly want to update the core development buildout of Plone: remove the package from `checkouts.cfg` and update its version pin (plus the pip counterparts), and do this in all relevant branches (6.0, 6.1, 6.2 where applicable).  We are using `plone.releaser` as wrapper around `zest.releaser` to do this automatically when you do such a release.

dataflake:
> > Would it make sense to add a PyPI publish step to the default template?
> 
> We could switch all packages to trusted publishing, this would allow people who are not listed on PyPI as maintainers to create releases, they'll then just need the approval of one of the maintainers. I am not sure that this is a use case.

No, I didn't have use cases in mind where non-PyPI-maintainers can publish packages. My only thought was why stop with the c-code templates, we could make it the same process everywhere so it's consistent, and it's supposedly safer. Supply chain security is becoming a more and more important issue.

> If you are not on PyPI as maintainer, then likely you don't have the rights to push directly to master/main either. So you would have to both create the tag for the release, and make a PR to merge the release commits. Then I think it is easier for everyone to let a maintainer do it all.

People who are not PyPI maintainers cannot publish straight to PyPI now and they shouldn't in the future. I have not tested the use case for those who can perform all the steps on GitHub but don't have PyPI publishing rights. I don't know what happens if they (or another GitHub admin) clicks OK to allow the "deployment" to PyPI.

icemac:
> People who are not PyPI maintainers cannot publish straight to PyPI now and they shouldn't in the future. I have not tested the use case for those who can perform all the steps on GitHub but don't have PyPI publishing rights. I don't know what happens if they (or another GitHub admin) clicks OK to allow the "deployment" to PyPI.

I'd expect that the package is then deployed to PyPI because trusted publishing is detached from the user.
The `Security history` of `zope.interface` says about the lately added sdist and wheels: `Added by: OpenID created token` and this token points to the trusted publisher management where the GitHub `tests.yml` workflow is allowed to create releases. There is no connection to the GitHub user anymore. That's why there is this confirmation email before anything is published.

dataflake:
Here's a quick list of positives and negatives, do you have any additional points that would fit in?

Positive:
- less work and less chances to make mistakes, you don't have to create and publish sdists/wheels from the command line
- (supposedly) safer than manual publishing
- all admins get emailed for all releases (no release goes unnoticed)

Negative
- less fine-grained access control, all release managers/admins can now publish
- all admins get emailed for all releases (annoying for packages you don't care about)

@mauritsvanrees Wouldn't your additional automation still work? It seems you'd just have to tell plone.releaser/zest.releaser to skip PyPI publishing.

mauritsvanrees:
Yes, `zest.releaser` should still work fine.

It could be annoying to get all these emails, and then having to click them, only to see that someone has already approved.  But we are not making dozens of releases per day, so that should be okay.  I guess you can also configure you email client to ignore these (remove or mark as read) and then you probably see them in the online GitHub notifications.

For pure Python packages, it would be more work than before: you have this extra approval step.  But it's not much.

dataflake:
While working on https://github.com/zopefoundation/meta/pull/418 I got a better picture of the release approval process. You can change the people and/or teams that get email to approve a publishing deployment by going to Settings -> Environments in your repository and then clicking on the `pypi` environment. In theory you could remove all of these approvers, which presumably means only admins can approve as long as you check the box about admins being able to bypass that rule.

PR #373, opened 2026-05-12
Fix tests for Python 3.15.

icemac:
It has a different repr for some cases.

See https://github.com/zopefoundation/zope.interface/actions/runs/25628471664/job/75227691953 for a failed test.


dataflake:
This seems to be using outdated zope.meta templates and is hitting the bad cache hit issues.

icemac:
@dataflake I updated with the current meta version, please re-review.

dataflake (approved):

icemac:
Thank you!

PR #377, opened 2026-07-02
Add opt-in strict mode for ambiguous utility lookups

niz11:
## Problem

When two utilities are registered for *sibling* subinterfaces of a queried
interface (neither extends the other), `getUtility`/`queryUtility` return a
single "best match" chosen by registration order. Adding a second utility can
therefore silently change which one is returned, with no error:

```python
class IParent(Interface): pass
class IChildA(IParent): pass
class IChildB(IParent): pass

provideUtility(A(), IChildA)
provideUtility(B(), IChildB)
getUtility(IParent)   # -> whichever was registered last; no error
```

## Change

Add an **opt-in** `Components.strict` flag (default `False`). When set,
`getUtility`/`queryUtility` raise the new `AmbiguousUtilityLookupError`
(a `ComponentLookupError` subclass) when the requested interface is matched by
two or more mutually-incomparable registrations, instead of returning an
order-dependent result. Also adds `Components.findAmbiguousUtilities()` to
audit a registry for such cases without changing behaviour.

- **Non-breaking:** default behaviour is unchanged (last registration wins).
- **No `IComponents` change** and **no C-extension change** (pure Python in `registry.py`).
- Exact matches and more-general subinterfaces still resolve deterministically;
  only genuinely incomparable ties are flagged.
- The check is O(registrations) per lookup and uncached, so `strict` is aimed at
  development/test/start-up use.

## Open questions for reviewers

1. `queryUtility` currently *raises* in strict mode too (ambiguity treated as a
   config error, distinct from not-found -> `default`). Would you prefer it
   return `default` instead and reserve raising for `getUtility`?
2. Worth caching the check (invalidated on registry change) so `strict` is
   viable outside dev/test?

## Tests

`test_registry.py` covers siblings-raise, exact-match, most-general-wins,
per-name isolation, across-bases, diamond bases, default-stays-silent, and the
audit. Full suite passes.


src/zope/interface/registry.py
```diff
@@ -294,15 +308,111 @@ def registeredUtilities(self):
...
+        the result would otherwise depend on registration order.
+        """
+        if self.strict:
+            self._check_ambiguous_utility(provided, name)
         utility = self.utilities.lookup((), provided, name)
         if utility is None:
             raise ComponentLookupError(provided, name)
         return utility
 
+    def findAmbiguousUtilities(self):
```
icemac, line 387: Calling this method might return many false-positives because having this ambiguity only triggers to be a problem if a utility for the parent interface is requested. But would be interesting to see what it returns on a sizable registry.

icemac (commented):
LGTM.

icemac:
Thank you for your contribution.

According to the contributing policies of the zopefoundation organization you need to sign a contributor agreement before any non-trivial change can be merged. For details please consult the [Contributing guidelines for zopefoundation projects](https://github.com/zopefoundation/.github/blob/master/CONTRIBUTING.md).

PR #378, opened 2026-08-10
Upgrade pip via `python -m pip` so the PyPy/Windows job stops failing

icemac:
On Windows the `pip.exe` shim refuses to replace itself (`ERROR: To modify pip, please run
the following command: ...python.exe -m pip install -U pip`), and a multi-line `pwsh`
`run:` only propagates the last command's exit code, so that failure went unnoticed. The
PyPy/Windows job therefore stayed on the pip 24.0 that `ensurepip` ships, and pip 24.0
does not invalidate corrupt entries in its HTTP cache — which we persist via
`actions/cache`. Result: `build-package (pypy-3.11, windows-latest)` failed with
`OSError: Connection broken: IncompleteRead(...)` on every run that restored the bad
entry, until GitHub happened to evict it.

Verified over two runs: pip now reaches 26.2.1, and the job passes on a **cache hit** of
the freshly written entry — the condition that failed nine times since May. The stale
cache entry was deleted, so `master` is unblocked regardless of this PR.

Next step: I plan to put this into `zope.meta` — `tests.yml` is generated, so this fix
would otherwise be reverted on the next template run. See zopefoundation/meta#436.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -163,12 +163,12 @@ jobs:
       - name: Install Build Dependencies (3.15)
         if: startsWith(matrix.python-version, '3.15')
         run: |
-          pip install -U pip
+          python -m pip install -U pip
           pip install -U "setuptools >= 78.1.1,< 82" wheel twine cffi pycparser
       - name: Install Build Dependencies
         if: ${{ !startsWith(matrix.python-version, '3.15') }}
         run: |
-          pip install -U pip
+          python -m pip install -U pip
           pip install -U "setuptools >= 78.1.1,< 82" wheel twine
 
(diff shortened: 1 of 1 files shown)


icemac:
Verified over two runs: [31365469354](https://github.com/zopefoundation/zope.interface/actions/runs/31365469354) (cache miss, writes a fresh cache) and [31365967253](https://github.com/zopefoundation/zope.interface/actions/runs/31365967253) (cache hit on that entry — the condition that failed nine times since May). Both fully green: pip reaches 26.2.1, no `IncompleteRead`, no `ERROR: To modify pip`.

— _Comment created by Claude_

jensens (approved):

PR #379, opened 2026-08-12
Configuring for c-code

icemac:
Applies the templates from zopefoundation/meta#437 (not merged yet, hence the
`commit-id` pointing at that branch).

The point of this PR is the Windows CI evidence #437 cannot get on its own:
every `run:` block now uses `bash`, so a failing command no longer gets
swallowed on Windows, and the OS-split pip cache steps collapse into one pair.

`pyproject.toml` loses the `zope.meta` header comment — that is `tomlkit`
round-tripping in `config-package`, unrelated to #437.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -21,8 +21,8 @@ jobs:
     name: linting
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
-    - uses: actions/setup-python@v6
+    - uses: actions/checkout@v7
+    - uses: actions/setup-python@v7
       with:
         python-version: '3.13'
     - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd  #v3.0.1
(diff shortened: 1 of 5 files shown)


icemac:
The previous run reproduced the original failure on `build-package (pypy-3.11,
windows-latest)` — on a **cache hit**, with pip 26.2.1 already installed:

```
Cache restored from key: Windows-X64-pip-pypy-3.11
...
ERROR: Could not install packages due to an OSError:
  ('Connection broken: IncompleteRead(20843 bytes read, 276 more expected)', ...)
```

So upgrading pip does not heal a corrupt cache entry; the never-varying key does the damage.
`actions/cache` states it plainly: "if the provided `key` matches an existing cache, a new
cache is not created". The entry is immortal until GitHub evicts it.

Latest commit adds a weekly epoch to the key, so a poisoned entry is abandoned within a week
instead of failing every run. `restore-keys` is deliberately not reintroduced (see
zopefoundation/meta@f73efb4) — the fresh key starts cold rather than copying the bad content
forward. Template side: zopefoundation/meta#437.

The branch name still says `764de66b` because I kept the existing branch to avoid
force-pushing; `.meta.toml` points at the current template commit.

— _Comment created by Claude_

dataflake (approved):

icemac:

Issue #380, opened 2026-08-13
copt: LookupBase cache dicts are read as borrowed pointers and can be freed under the reader on free-threaded builds


espressolee:
This was reported through the project's security address on 2026-08-07 and is being filed
publicly at the maintainer's request.

**This concerns the free-threaded CPython build** (`Py_GIL_DISABLED`, PEP 703).

This issue reports the defect and three open questions. **It deliberately does not carry a
patch.** The questions below are about CPython's free-threaded GC and synchronization
contracts, and I cannot settle them from outside the project; proposing a synchronization
design past them would be presenting an unverified scheme as a ready fix. A concrete change
can follow in a PR once they are decided — and after I have signed the contributor agreement
Jens pointed me at.

**What I am asking for:** a decision on the three open questions at the end, before any
implementation starts.

### Problem

The `LookupBase` cache fast paths load `self->_cache` / `_mcache` / `_scache` and pass the
loaded pointer into `_subcache`, which uses it. No reference is taken on the loaded object.
Concurrently `LookupBase.changed()` runs `LB_changed` → `LB_clear` →
`Py_CLEAR(self->_cache)`, which drops the object's only strong reference to that dict and
frees it. The interval between the load and the use is unprotected.

The three sites (tag 8.5; line numbers identical on master):

```
:1259  ASSURE_DICT(self->_cache);                    in _getcache
:1261  cache = _subcache(self->_cache, provided);
:1617  ASSURE_DICT(self->_mcache);                   in _lookupAll
:1619  cache = _subcache(self->_mcache, provided);
:1702  ASSURE_DICT(self->_scache);                   in _subscriptions
:1704  cache = _subcache(self->_scache, provided);
```

`ASSURE_DICT` (:67) creates the dict only when the field is NULL and takes no reference.

**This is the outer half of a hazard whose inner half is already closed.** The comment above
`_subcache` (:1222–1224) states it:

> Uses PyDict_GetItemRef() for a strong reference, needed for free-threaded Python where
> another thread could clear the parent cache between PyDict_GetItem()'s return and our use
> of the result.

PR #352 did that for the **values** fetched out of the cache dict. The cache dict **itself**
is still loaded from the instance with no reference taken, and the same `changed()` that #352
was defending against frees exactly that pointer.

### Every access to the three fields

Enumerated because the scope of any fix depends on it. Nothing below is labelled safe or
protected — these are the operations as the current source performs them.

| line | operation | kind |
|---|---|---|
[...]

PR #381, opened 2026-09-01
Fix SystemError from NULL-without-exception returns in C optimizations

### LINKED ISSUE #359: copt: NULL-without-exception returns in SB_extends, IB__adapt__, CPB_descr_get
A detailed analysis of `_zope_interface_coptimizations.c` was published at
https://gist.github.com/devdanzin/5afbad864934b165a2cc080f110aa720 by @devdanzin.
We reviewed the report with agent support and verified the findings against the
current source (zope.interface 8.2). This issue is one of several extracted from
that review.

### Problem

When `_implied` or `_cls` fields are NULL (possible because they use
`T_OBJECT_EX` which allows deletion), three functions return NULL without setting
a Python exception, causing `SystemError`:

- `SB_extends` line 331: `if (implied == NULL) return NULL;`
- `IB__adapt__` line 789: same pattern for `_implied`
- `CPB_descr_get` line 618: `if (self->_cls == NULL) return NULL;`

### Reproducer

```python
from zope.interface._zope_interface_coptimizations import SpecificationBase

sb = SpecificationBase()
sb.isOrExtends(Interface)
# SystemError: <method 'isOrExtends' of 'SpecificationBase' objects>
#   returned NULL without setting an exception
```

```python
from zope.interface._zope_interface_coptimizations import ClassProvidesBase

cpb = ClassProvidesBase.__new__(ClassProvidesBase)
cpb.__get__(object(), type)
# SystemError: <slot wrapper '__get__' of 'ClassProvidesBase' objects>
#   returned NULL without setting an exception
```

### Fix

Set an exception before returning NULL:

```c
if (implied == NULL) {
    PyErr_SetString(PyExc_AttributeError, "_implied");
    return NULL;
}
```

Same for `_cls` in `CPB_descr_get`.

### Scope

3 sites. No behavioral change for properly initialized objects.

### Tests

Directly testable. The reproducers from the report work as-is as test cases:

```python
def test_isOrExtends_raises_AttributeError_when_implied_is_None(self):
    from zope.interface.interface import SpecificationBase
    sb = SpecificationBase()
    # _implied is not set (NULL in C)
    with self.assertRaises(AttributeError):
        sb.isOrExtends(object())

def test_CPB_descr_get_raises_AttributeError_when_cls_is_None(self):
    from zope.interface._zope_interface_coptimizations import ClassProvidesBase
    cpb = ClassProvidesBase.__new__(ClassProvidesBase)
    with self.assertRaises(AttributeError):
        cpb.__get__(object(), type)
```

These currently raise `SystemError` (the bug). After the fix they raise
`AttributeError`. The tests verify the fix works.

Note: these test the C implementation specifically. The Python fallback may
behave differently (regular `AttributeError` from Python attribute access). Both
paths should be tested via the existing Fallback/Optimized test class pattern.

### Performance impact

Zero. The NULL check already exists in the current code. The fix only adds a
`PyErr_SetString` call inside the existing never-taken branch. No new branches,
no new instructions on the happy path.

afonsojanu:
Fixes #359.

The gist linked from that issue (https://gist.github.com/devdanzin/5afbad864934b165a2cc080f110aa720) flags three spots in `_zope_interface_coptimizations.c` that return `NULL` without setting an exception when a field is unset:

- `SB_extends` (`isOrExtends`), when `self->_implied` is `NULL`
- `IB__adapt__`, when the provider's `_implied` is `NULL`
- `CPB_descr_get` (`__get__`), when `self->_cls` is `NULL`

Both `_implied` and `_cls` are declared `T_OBJECT_EX`, which means they can legitimately be `NULL`, either because they were never assigned or because they were `del`eted. Returning `NULL` from a C function without setting an exception is undefined behavior from CPython's point of view, and in practice it surfaces to Python callers as a fairly unhelpful `SystemError` rather than the `AttributeError` you'd normally expect from accessing a missing attribute.

I confirmed all three with the reproducers from the issue before touching anything:

```
>>> from zope.interface._zope_interface_coptimizations import SpecificationBase
>>> sb = SpecificationBase()
>>> sb.isOrExtends(Interface)
SystemError: <method 'isOrExtends' of ...> returned NULL without setting an exception
```

The fix is exactly what the issue suggests: call `PyErr_SetString(PyExc_AttributeError, ...)` before each of the three `return NULL` sites. No behavioral change for properly-initialized objects, since these branches were already unreachable in that case.

I also added regression tests for all three (`test_interface.py`, `test_declarations.py`), verified against the unpatched C source that they fail with the original `SystemError`, and confirmed the Python fallback implementation (`SpecificationBasePy`) already raises `AttributeError` correctly, so this brings the C path in line with it. Full local test suite (1377 tests) passes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,10 +5,11 @@ Change log
 ----------------
 
 - Fix the C optimizations returning ``NULL`` without setting an exception
-  in three spots (``SpecificationBase.isOrExtends``,
-  ``InterfaceBase.__adapt__``, and ``ClassProvidesBase.__get__``) when
-  ``_implied`` or ``_cls`` had not been set, e.g. after ``del``. This used
-  to surface to callers as ``SystemError`` instead of ``AttributeError``.
+  in four spots (``SpecificationBase.isOrExtends``,
+  ``InterfaceBase.__adapt__``, and both the ``_cls`` and ``_implements``
+  accesses in ``ClassProvidesBase.__get__``) when the corresponding
+  attribute had not been set, e.g. after ``del``. This used to surface to
+  callers as ``SystemError`` instead of ``AttributeError``.
   See `issue 359 <https://github.com/zopefoundation/zope.interface/issues/359>`_.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@ Change log
 8.7 (unreleased)
 ----------------
 
+- Fix the C optimizations returning ``NULL`` without setting an exception
+  in three spots (``SpecificationBase.isOrExtends``,
+  ``InterfaceBase.__adapt__``, and ``ClassProvidesBase.__get__``) when
+  ``_implied`` or ``_cls`` had not been set, e.g. after ``del``. This used
+  to surface to callers as ``SystemError`` instead of ``AttributeError``.
+  See `issue 359 <https://github.com/zopefoundation/zope.interface/issues/359>`_.
+
 
 8.6 (2026-08-20)
 ----------------
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -330,6 +330,7 @@ SB_extends(SB* self, PyObject* other)
 
     implied = self->_implied;
     if (implied == NULL) {
+        PyErr_SetString(PyExc_AttributeError, "_implied");
         return NULL;
     }
 
@@ -619,8 +620,10 @@ CPB_descr_get(CPB* self, PyObject* inst, PyObject* cls)
 {
     PyObject* implements;
 
-    if (self->_cls == NULL)
+    if (self->_cls == NULL) {
+        PyErr_SetString(PyExc_AttributeError, "_cls");
         return NULL;
+    }
 
     if (cls == self->_cls) {
         if (inst == NULL) {
@@ -792,6 +795,7 @@ IB__adapt__(PyObject* self, PyObject* obj)
 
         implied = ((SB*)decl)->_implied;
         if (implied == NULL) {
+            PyErr_SetString(PyExc_AttributeError, "_implied");
             Py_DECREF(decl);
             return NULL;
         }
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -632,7 +632,11 @@ CPB_descr_get(CPB* self, PyObject* inst, PyObject* cls)
         }
 
         implements = self->_implements;
-        Py_XINCREF(implements);
+        if (implements == NULL) {
+            PyErr_SetString(PyExc_AttributeError, "_implements");
+            return NULL;
+        }
+        Py_INCREF(implements);
         return implements;
     }
 
--- a/src/zope/interface/tests/test_declarations.py
+++ b/src/zope/interface/tests/test_declarations.py
@@ -1791,6 +1791,20 @@ class ClassProvidesBaseFallbackTests(unittest.TestCase):
         cpb = klass.__new__(klass)
         self.assertRaises(AttributeError, cpb.__get__, object(), type)
 
+    def test___get___raises_AttributeError_when_implements_not_set(self):
+        # https://github.com/zopefoundation/zope.interface/issues/359
+        # Same underlying bug as the _cls case above, but for the
+        # _implements slot: the C implementation returned NULL without
+        # setting an exception once _cls matched and inst was not None,
+        # surfacing as a SystemError instead of AttributeError.
+        class Derived(self._getTargetClass()):
+            def __init__(self, k):
+                self._cls = k
+
+        klass = object
+        cpb = Derived(klass)
+        self.assertRaises(AttributeError, cpb.__get__, object(), klass)
+
 
 class ClassProvidesBaseTests(
     OptimizationTestMixin,
--- a/src/zope/interface/tests/test_declarations.py
+++ b/src/zope/interface/tests/test_declarations.py
@@ -1782,6 +1782,15 @@ class ClassProvidesBaseFallbackTests(unittest.TestCase):
         self.assertRaises(AttributeError, getattr, Bar, '__provides__')
         self.assertRaises(AttributeError, getattr, bar, '__provides__')
 
+    def test___get___raises_AttributeError_when_cls_not_set(self):
+        # https://github.com/zopefoundation/zope.interface/issues/359
+        # The C implementation used to return NULL without setting an
+        # exception in this case, which surfaces to callers as a
+        # SystemError instead of the expected AttributeError.
+        klass = self._getTargetClass()
+        cpb = klass.__new__(klass)
+        self.assertRaises(AttributeError, cpb.__get__, object(), type)
+
 
 class ClassProvidesBaseTests(
     OptimizationTestMixin,
--- a/src/zope/interface/tests/test_interface.py
+++ b/src/zope/interface/tests/test_interface.py
@@ -243,6 +243,16 @@ class GenericSpecificationBaseTests(unittest.TestCase):
         with self.assertRaises(MemoryError):
             IFoo.isOrExtends(BadHash())
 
+    def test_isOrExtends_raises_AttributeError_when_implied_not_set(self):
+        # https://github.com/zopefoundation/zope.interface/issues/359
+        # The C implementation used to return NULL without setting an
+        # exception in this case, which surfaces to callers as a
+        # SystemError instead of the expected AttributeError.
+        klass = self._getTargetClass()
+        sb = klass.__new__(klass)
+        with self.assertRaises(AttributeError):
+            sb.isOrExtends(object())
+
 
 class SpecificationBaseTests(
     GenericSpecificationBaseTests,
@@ -535,6 +545,29 @@ class InterfaceBasePyTests(InterfaceBaseTestsMixin, unittest.TestCase):
         with _Monkey(interface, adapter_hooks=[_hook_miss, _hook_hit]):
             self.assertIs(ib.__adapt__(adapted), adapted)
             self.assertEqual(_missed, [(ib, adapted)])
+
+
+class Test__adapt__(unittest.TestCase):
+    # https://github.com/zopefoundation/zope.interface/issues/359
+    # If ``providedBy(obj)`` is a SpecificationBase instance whose
+    # ``_implied`` was never populated, ``__adapt__`` used to return
+    # NULL without setting an exception in the C implementation,
+    # surfacing to callers as SystemError instead of AttributeError.
+
+    def test___adapt___raises_AttributeError_when_providers_implied_not_set(
+        self,
+    ):
+        from zope.interface import Interface
+        from zope.interface.interface import SpecificationBase
+
+        class Provider:
+            pass
+
+        provider = Provider()
+        provider.__providedBy__ = SpecificationBase()
+
+        with self.assertRaises(AttributeError):
+            Interface.__adapt__(provider)
 
 
 class SpecificationTests(unittest.TestCase):
--- a/src/zope/interface/tests/test_interface.py
+++ b/src/zope/interface/tests/test_interface.py
@@ -547,27 +547,59 @@ class InterfaceBasePyTests(InterfaceBaseTestsMixin, unittest.TestCase):
             self.assertEqual(_missed, [(ib, adapted)])
 
 
-class Test__adapt__(unittest.TestCase):
+class GenericTest__adapt__(unittest.TestCase):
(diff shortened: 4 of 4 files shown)


icemac (commented):
Thanks for tracking these down — the three C fixes match the Python fallbacks and are correct. I built the branch and ran the tests in both C and pure-Python mode; two things need attention before this can merge.

### 1. The new `__adapt__` test fails under `PURE_PYTHON=1`

`Test__adapt__.test___adapt___raises_AttributeError_when_providers_implied_not_set` (`src/zope/interface/tests/test_interface.py:550`) is a plain `TestCase` with no C/fallback pairing, so it exercises whichever implementation happens to be active — and the two behave differently here:

* C `providedBy()` accepts the spec via `PyObject_TypeCheck(result, specification_base_class)`, so it hands the bare `SpecificationBase()` straight to `__adapt__`, whose `_implied` is unset → `AttributeError`. Test passes.
* Python `providedBy()` (`declarations.py:1126-1131`) instead probes `r.extends`. The Python `SpecificationBase` has no `extends` — that lives on `Specification` — so it raises `AttributeError` internally and falls through to `implementedBy(Provider)`. That spec's `_implied` *does* contain `Interface`, so `__adapt__` returns `provider` and nothing is raised.

```console
$ PURE_PYTHON=1 python -m unittest zope.interface.tests.test_interface.Test__adapt__
FAIL: ... AssertionError: AttributeError not raised
```

This breaks `tox -e pure` (`tox.ini:27`, `tox.ini:50`), the `PURE_PYTHON: 1` job at `.github/workflows/tests.yml:386`, and every PyPy run.

Rather than skipping the test on pure Python, could you restructure it along the lines of the mixin pattern the other two new tests use (`GenericSpecificationBaseTests` / `ClassProvidesBaseFallbackTests`)? That way both implementations are actually exercised instead of one or the other.

### 2. `CPB_descr_get` still has the same bug for `_implements`

`src/zope/interface/_zope_interface_coptimizations.c:634`: the new `self->_cls == NULL` guard is right, but a few lines further down:

```c
implements = self->_implements;
Py_XINCREF(implements);
return implements;
```

`_implements` is also a `T_OBJECT_EX` member (see `CPB_members`), so it can be `NULL` when never assigned or after `del` — the `Py_XINCREF` acknowledges as much — and returning it raises the identical `SystemError`. Reproduced against this branch:

```pycon
>>> class Derived(ClassProvidesBase):
...     def __init__(self, k): self._cls = k      # _implements left unset
>>> class Foo: pass
>>> Foo.__provides__ = Derived(Foo)
>>> Foo().__provides__
SystemError: error return without exception set
```

The Python fallback raises `AttributeError: 'Derived' object has no attribute '_implements'` for the same input, so the C/Python divergence this PR sets out to close is still open here. It needs the same `PyErr_SetString(PyExc_AttributeError, "_implements"); return NULL;` treatment, ideally with a test in `ClassProvidesBaseFallbackTests`.

### Checked and fine

* `test_isOrExtends_raises_AttributeError_when_implied_not_set` and `test___get___raises_AttributeError_when_cls_not_set` pass in both C and pure-Python mode.
* No other `T_OBJECT_EX` member is dereferenced-and-returned elsewhere; `SB_providedBy`, `SB_implementedBy` and `SB__call__` all route through the now-fixed `SB_extends`.
* Setting the exception before `Py_DECREF(decl)` in `IB__adapt__` is safe — CPython's finalizer and weakref-callback machinery save and restore the pending exception.
* The `CHANGES.rst` entry is accurate.

— _Comment created by Claude_

dataflake:
Please sign the contributor agreement so we can accept this patch, you can do that electronically by following the link Contributor Agreement Form for Zope on the [Plone Foundation contributor agreement page](https://plone.org/foundation/contributors-agreement).

afonsojanu:
Following up on the CI failure from a couple of days ago — it turns out this wasn't a one-off flake on a single Windows free-threaded leg after all. It's now failing consistently on essentially every leg (Linux, macOS, Windows, every Python version 3.10 through 3.15), always the same test: test___adapt___raises_AttributeError_when_providers_implied_not_set.

I've dug into this properly now rather than just retriggering. A few things I've confirmed:

- master's own CI at the commit this branch is based on (2bd6aec) is fully green, so this is genuinely introduced by this branch's diff, not a pre-existing infra issue.
- The actual diff touching the C code (b4fbe4f) is three lines, each just adding a PyErr_SetString call before an existing `return NULL`. I've read through it carefully and can't find a correctness issue in it.
- SB_type_def uses PyBaseObject_Type.tp_new, so a freshly constructed SpecificationBase() should always get a zero-initialized struct via the normal allocator, ruling out the freelist/memory-reuse theory I was chasing before.
- I ran the failing test, its whole test file, and the full `python -m unittest discover -s src/zope` suite 20 times in a row locally (macOS, Python 3.11, C extension built and loaded) and it passed every single time with zero flakiness.

So I'm stuck on reproducing this locally despite it failing reliably on every CI platform. My best guess at this point is some interaction with test collection order that differs between my local run and CI's, but I haven't been able to pin down which earlier test would need to run first to trigger it, and I don't want to guess at a fix for something I can't actually reproduce and verify.

Happy to keep digging if there's a specific angle you'd suggest, or if you have any insight into how test discovery order might differ on CI versus a local run of the same command.

tseaver:
@afonsojanu Did you see the review by @icemac? The failure is in the "Run tests without C extensions" section, which matches the scope of his review pretty well (play along at home by running with `PURE_PYTHON=1`).

afonsojanu:
Both fixed, thanks for the thorough read.

CPB_descr_get: same NULL-without-exception bug for _implements, same treatment as the other three spots (check for NULL, raise AttributeError, return NULL). Added a test mirroring the existing _cls one.

The __adapt__ test: split it into a base class plus a subclass each for pure-Python and C, matching the mixin pattern already used elsewhere in this file. Along the way I confirmed exactly what you described: a bare SpecificationBasePy() genuinely can't reach the buggy path at all, since providedBy()'s Python fallback only trusts it as a specification if it has .extends, which lives on Specification, not the bare base class. Without that, it falls straight through to implementedBy(Provider) and the bug never gets exercised. I switched the Python-side construction to Specification.__new__(Specification), which has extends but still leaves _implied unset the same way, and confirmed both sides fail without the fix and pass with it under PURE_PYTHON=1 and normal runs.

Full suite (1380 tests) green both ways. Updated CHANGES.rst to mention four spots instead of three.

icemac:
> Please sign the contributor agreement so we can accept this patch, you can do that electronically by following the link Contributor Agreement Form for Zope on the [Plone Foundation contributor agreement page](https://plone.org/foundation/contributors-agreement).

@afonsojanu This still seems to be open.

jensens (commented):
Review of c490f82, on top of the existing one; overlaps marked ↔.

### Summary

- The four fixes are correct. New tests pass, full suite green: 1376 tests, C and `PURE_PYTHON=1` on 3.14.3, C on 3.14.2t. The failing `pypy-3.11 / windows-latest` job dies in `Install zope.interface` (network), same as on master dd65a10 on 2026-09-20 — not this PR.
- 1–2: two more instances of the bug class the PR title names, in the touched functions, reproduced on c490f82. Small fixes; here or follow-up, maintainers' call.
- 3: the new `__adapt__` test pair exercises only the C path in a C run.
- 4–6: optional.

### Same bug class, still open

**1. Non-dict `_implied` segfaults instead of raising** — [`SB_extends` L337][c337], [`IB__adapt__` L807][c807]

- `_implied` is a writable `T_OBJECT_EX` member; the PR guards `NULL` but `PyDict_Contains()` still runs on whatever is there.
- `sb = SpecificationBase(); sb._implied = []; sb.isOrExtends(object())` → SIGSEGV. Same via `Interface.__adapt__(obj)` with `obj.__providedBy__ = Specification.__new__(Specification)` and `_implied = []`. Reproduced on 3.14.3 and 3.14.2t.
- Python fallback returns `False` for the same input.
- Fix: `PyDict_Check()` → `TypeError`, once, in a small helper shared by both sites.

**2. `IB_richcompare` with unset `__name__` / `__module__` → `SystemError`** — [L1049–L1059][c1049]

- `InterfaceBase.__new__(InterfaceBase) == InterfaceBase('IFoo', 'mod')` → `SystemError: Objects/object.c:1101: bad argument to internal function`. Also with the unset object on the right-hand side (`othername = otherib->__name__` at L1031). Reproduced on c490f82.
- Same trigger as #359 (`T_OBJECT_EX` slot never set or deleted); [`IB__hash__` L963–L967][c963] already guards both members.
- ↔ earlier review, "no other `T_OBJECT_EX` member is dereferenced-and-returned": these are passed to `PyObject_RichCompareBool()`, not returned, with the same effect.

### Tests

**3. `GenericTest__adapt__` / `Test__adapt__` both hit the C path in a C run** — [L550–L603][t550]

- `Specification` subclasses the active `SpecificationBase`, so in a C run `Specification.__new__(Specification)` is a C instance and `Interface.__adapt__` is the C method for both classes. Verified: with `InterfaceBasePy.__adapt__` instrumented, both tests pass with 0 calls into it. Under `PURE_PYTHON=1` the two classes are identical.
- The comment at L560 ("runs standalone against the pure-Python path") describes something that does not happen.
- The file's existing pattern covers both in one run: [`InterfaceBaseTestsMixin` L449][t449] with `InterfaceBaseTests` (C) and `InterfaceBasePyTests` (Python). One test there, `self._getTargetClass()('IFoo', 'mod').__adapt__(provider)` with `provider.__providedBy__ = Specification.__new__(Specification)`, raises `AttributeError` on both classes in a C run (checked), and replaces the two-class construct.

### Optional

**4. `_should_attempt_c_optimizations()` as the gate** — [L599][t599]

- Returns `True` whenever `PURE_PYTHON` is unset on CPython, even when the extension failed to import; `SpecificationBase` is then `SpecificationBasePy`, has no `extends`, and the test fails for the reason the class comment describes. Class identity (`SpecificationBase is not SpecificationBasePy`) is what the rest of the suite keys on. Moot after 3.

**5. `AttributeError` shape differs from the fallback** — [L333][c333], [L624][c624], [L636][c636], [L802][c802]

- C: `str(e) == '_implied'`, `e.name is None`, `e.obj is None`. Fallback and the member getter itself: `"'…SpecificationBase' object has no attribute '_implied'"`, `e.name == '_implied'`, `e.obj` set.
- The bare name follows [`IB__hash__`][c963], so consistent with the file. If parity with the fallback is wanted: `return PyObject_GetAttr(OBJECT(self), str__implied)` on the `NULL` path lets the member descriptor raise the canonical error.

**6. Borrowed `_implied` across `PyDict_Contains()`** — [L331–L337][c337]

- No strong reference is taken; `other.__hash__` can run Python inside the call. On a free-threaded build, `del sb._implied` from another thread in that window frees the dict under the reader. Same family as #380 / #384, only reachable by explicit slot mutation → out of scope here, noted for the follow-up on those.

---

- Method: automated `/code-review` pass in a sub-agent (diff + surrounding source at c490f82), then a verification pass in the main session, including reading the existing review and thread.
- Environment: Linux x86_64, gcc 13, CPython 3.14.3 (GIL) and 3.14.2t. Baseline: master dd65a10.
- Executed: 1, 2 and 5 with scripts on both interpreters; the instrumented run behind 3; new tests and full suite, C and pure.
- Code reading only: 4, 6.
- Proposed changes not implemented or tested.

[c333]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L333
[c337]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L331-L337
[c624]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L624
[c636]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L636
[c802]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L802
[c807]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L800-L807
[c963]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L963-L967
[c1049]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/_zope_interface_coptimizations.c#L1049-L1059
[t449]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/tests/test_interface.py#L449
[t550]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/tests/test_interface.py#L550-L603
[t599]: https://github.com/zopefoundation/zope.interface/blob/c490f82/src/zope/interface/tests/test_interface.py#L599

afonsojanu:
Just signed the Plone Foundation contributor agreement, should be good to go now.

PR #382, opened 2026-09-04
Fix free-threaded data race in the C LookupBase cache (#380)

### LINKED ISSUE #380: copt: LookupBase cache dicts are read as borrowed pointers and can be freed under the reader on free-threaded builds

espressolee:
Fixes #380.

This PR protects the lifetimes of the C `LookupBase` cache fields and the
`VerifyingBase` verification snapshot on free-threaded builds.

The original cache-field race occurs when a lookup initializes or retains
`_cache`, `_mcache`, or `_scache` while `changed()` concurrently detaches the
same field.  That can corrupt memory and terminate the interpreter with either
`PyMutex_Unlock: unlocking mutex that is not locked` or a segmentation fault.

`VerifyingBase` has the same cache fields plus a second lifetime boundary:
`_verify()` used `_verify_ro` and `_verify_generations` through Python callbacks
while `verify_changed()` could clear and replace those raw fields.

At the current head:

- cache dictionaries are allocated outside the object critical section,
  installed or retained while locked, and returned with a strong reference;
- `LookupBase.changed()` and `VerifyingBase.changed()` detach cache fields while
  locked and decref them after unlocking;
- `VerifyingBase.changed()` replaces `_verify_ro` and `_verify_generations` as
  one locked pair; and
- `_verify()` acquires one strong pair snapshot before any attribute lookup or
  comparison can execute Python.

The pre-3.13 fallback remains a no-op block because the GIL serializes those
builds.  `LB_clear()` and `VB_clear()` remain the GC/deallocation paths; runtime
invalidation uses the synchronized methods above.  No allocation, Python
callback, or potentially re-entrant decref occurs while the object critical
section is held.

This is a memory-safety fix, not a claim of linearizable invalidation.  An
in-flight lookup may still finish against a cache generation that `changed()`
has already detached.

Local validation at exact head `5ec2eda9f9e93f8a2c09485684eb1344efcccfa9`:

- both new subprocess regressions fail on predecessor `5a2fdda` (one SIGSEGV;
  one deterministic verification-snapshot `SystemError`);
- the two-test free-threaded regression module passes in 10/10 fresh processes
  on CPython 3.14.6t with the GIL disabled;
- CPython 3.14.6t C-extension suite: 1,369 tests, 0 failures, 7 skipped;
- CPython 3.12.11 C and pure-Python suites: 1,373 tests each, 0 failures,
  9 skipped; and
- changed-file pre-commit checks and C syntax compilation against CPython 3.13
  and 3.11 headers pass.

### DIFF
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -64,13 +64,6 @@ _PyDict_GetItemRef(PyObject *p, PyObject *key, PyObject **result)
 #define PyDict_GetItemRef _PyDict_GetItemRef
 #endif
 
-#define ASSURE_DICT(N)                                                         \
-    if (N == NULL) {                                                           \
-        N = PyDict_New();                                                      \
-        if (N == NULL)                                                         \
-            return NULL;                                                       \
-    }
-
 /* Py_BEGIN/END_CRITICAL_SECTION arrived in 3.13.  Where it is unavailable the GIL
  * serializes everything, so the fallback is a plain block that takes no lock.
  * Guarded by capability (#ifndef), not version, so a backport is respected. */
@@ -1178,6 +1171,7 @@ LB_traverse(LB* self, visitproc visit, void* arg)
 static int
 LB_clear(LB* self)
 {
+    /* Runtime invalidation uses LB_changed(); this is for GC/deallocation. */
     Py_CLEAR(self->_cache);
     Py_CLEAR(self->_mcache);
     Py_CLEAR(self->_scache);
@@ -1205,10 +1199,7 @@ LB_dealloc(LB* self)
 static PyObject*
 LB_changed(LB* self, PyObject* ignored)
 {
-    /* Detach the three cache dicts under a critical section so this cannot race a
-     * concurrent lookup()'s field access (#380).  DECREF them AFTER releasing the
-     * section: a dict dealloc can run arbitrary Python (__del__), which must not
-     * execute while the lock is held. */
+    /* Detach while locked, then decref where finalizers may safely run. */
     PyObject *cache, *mcache, *scache;
     Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
     cache = self->_cache;   self->_cache = NULL;
@@ -1236,20 +1227,10 @@ LB_changed(LB* self, PyObject* ignored)
             cache = c
         return cache
 */
-/* Ensure *field is a dict and return a NEW strong reference to it, taking a critical
- * section on 'self' so the field's create/read cannot race a concurrent detach in
- * LB_changed() on a free-threaded build (#380 Q2/Q3).  NO Python runs inside the
- * section: Py_NewRef and pointer publication cannot re-enter.  Dict allocation is
- * performed outside the section, after a locked miss, then the field is checked
- * again before publication.  The returned strong reference keeps the dict alive
- * if LB_changed() detaches the field immediately afterward.  Returns NULL (error
- * set) only on allocation failure.
- *
- * NOTE: this guarantees memory safety, not linearizable invalidation -- a lookup that
- * retains a now-detached cache may still populate/return from that retired generation.
- * That is acceptable per the cache's semantics (a stale/duplicate cache entry, never a
- * use-after-free); strict invalidation would require generation checks with callback
- * and deadlock implications out of scope for this fix. */
+/* Return a strong cache-field reference.  Allocation and decref stay outside the
+ * critical section, where Python execution is permitted.  This guarantees memory
+ * safety, not linearizable invalidation: an in-flight lookup may finish against a
+ * cache generation that changed() has detached. */
 static PyObject*
 _assure_cache_field(LB* self, PyObject** field)
 {
@@ -1924,6 +1905,7 @@ VB_traverse(VB* self, visitproc visit, void* arg)
 static int
 VB_clear(VB* self)
 {
+    /* Runtime invalidation uses verify_changed(); this is for GC/deallocation. */
     Py_CLEAR(self->_verify_generations);
     Py_CLEAR(self->_verify_ro);
     return LB_clear((LB*)self);
@@ -1950,11 +1932,13 @@ VB_dealloc(VB* self)
 static PyObject*
 _generations_tuple(PyObject* ro)
 {
-    int i, l;
+    Py_ssize_t i, l;
     PyObject* generations;
 
     l = PyTuple_GET_SIZE(ro);
     generations = PyTuple_New(l);
+    if (generations == NULL)
+        return NULL;
     for (i = 0; i < l; i++) {
         PyObject* generation;
 
@@ -1971,9 +1955,26 @@ _generations_tuple(PyObject* ro)
 static PyObject*
 verify_changed(VB* self, PyObject* ignored)
 {
-    PyObject *t, *ro;
+    PyObject *cache, *mcache, *scache;
+    PyObject *old_ro, *old_generations;
+    PyObject *replaced_ro, *replaced_generations;
+    PyObject *t, *ro, *generations;
 
-    VB_clear(self);
+    /* Clear all lookup and verification fields as one state transition.  The
+     * detached objects remain alive until after the critical section. */
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    cache = self->lookup._cache;   self->lookup._cache = NULL;
+    mcache = self->lookup._mcache; self->lookup._mcache = NULL;
+    scache = self->lookup._scache; self->lookup._scache = NULL;
+    old_ro = self->_verify_ro;     self->_verify_ro = NULL;
+    old_generations = self->_verify_generations;
+    self->_verify_generations = NULL;
+    Py_END_CRITICAL_SECTION();
+    Py_XDECREF(cache);
+    Py_XDECREF(mcache);
+    Py_XDECREF(scache);
+    Py_XDECREF(old_ro);
+    Py_XDECREF(old_generations);
 
     t = PyObject_GetAttr(OBJECT(self), str_registry);
     if (t == NULL)
@@ -1994,16 +1995,24 @@ verify_changed(VB* self, PyObject* ignored)
     if (ro == NULL)
         return NULL;
 
-    self->_verify_generations = _generations_tuple(ro);
-    if (self->_verify_generations == NULL) {
+    generations = _generations_tuple(ro);
+    if (generations == NULL) {
         Py_DECREF(ro);
         return NULL;
     }
 
+    /* Another changed() may have published a newer pair while Python ran
+     * above.  Replace both fields atomically and retire that pair afterward. */
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    replaced_ro = self->_verify_ro;
+    replaced_generations = self->_verify_generations;
     self->_verify_ro = ro;
+    self->_verify_generations = generations;
+    Py_END_CRITICAL_SECTION();
+    Py_XDECREF(replaced_ro);
+    Py_XDECREF(replaced_generations);
 
-    Py_INCREF(Py_None);
-    return Py_None;
+    Py_RETURN_NONE;
 }
 
 /*
@@ -2015,11 +2024,25 @@ verify_changed(VB* self, PyObject* ignored)
 static int
 _verify(VB* self)
 {
+    PyObject* ro = NULL;
+    PyObject* generations = NULL;
     PyObject* changed_result;
 
+    /* Snapshot the pair under the same lock used by verify_changed().  Every
+     * subsequent borrowed item is then owned transitively by these references,
+     * even when PyObject_GetAttr() or comparison runs Python. */
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
     if (self->_verify_ro != NULL && self->_verify_generations != NULL) {
-        int i, l;
-        l = PyTuple_GET_SIZE(self->_verify_ro);
+        ro = Py_NewRef(self->_verify_ro);
+        generations = Py_NewRef(self->_verify_generations);
+    }
+    Py_END_CRITICAL_SECTION();
+
+    if (ro != NULL) {
+        Py_ssize_t i, l;
+        l = PyTuple_GET_SIZE(ro);
+        if (PyTuple_GET_SIZE(generations) != l)
+            goto changed;
 
         /* Compare each registry's current _generation counter against the
          * snapshot stored in _verify_generations, without allocating a
@@ -2028,23 +2051,34 @@ _verify(VB* self)
          * RichCompareBool.  This version compares in-place and exits
          * early on the first mismatch. */
         for (i = 0; i < l; i++) {
-            PyObject *reg = PyTuple_GET_ITEM(self->_verify_ro, i);
+            PyObject *reg = PyTuple_GET_ITEM(ro, i);
             PyObject *current_gen = PyObject_GetAttr(reg, str_generation);
-            if (current_gen == NULL)
+            if (current_gen == NULL) {
+                Py_DECREF(ro);
+                Py_DECREF(generations);
                 return -1;
+            }
 
             PyObject *stored_gen = PyTuple_GET_ITEM(
-                self->_verify_generations, i);
+                generations, i);
             int eq = PyObject_RichCompareBool(current_gen, stored_gen, Py_EQ);
             Py_DECREF(current_gen);
 
-            if (eq < 0) return -1;   /* error */
+            if (eq < 0) {
+                Py_DECREF(ro);
+                Py_DECREF(generations);
+                return -1;
+            }
             if (eq == 0) goto changed; /* mismatch — early exit */
         }
+        Py_DECREF(ro);
+        Py_DECREF(generations);
         return 0;  /* all match, cache is still valid */
     }
 
 changed:
+    Py_XDECREF(ro);
+    Py_XDECREF(generations);
     changed_result =
       PyObject_CallMethodObjArgs(OBJECT(self), strchanged, Py_None, NULL);
     if (changed_result == NULL)
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -71,6 +71,14 @@ _PyDict_GetItemRef(PyObject *p, PyObject *key, PyObject **result)
             return NULL;                                                       \
     }
 
+/* Py_BEGIN/END_CRITICAL_SECTION arrived in 3.13.  Where it is unavailable the GIL
+ * serializes everything, so the fallback is a plain block that takes no lock.
+ * Guarded by capability (#ifndef), not version, so a backport is respected. */
+#ifndef Py_BEGIN_CRITICAL_SECTION
+#define Py_BEGIN_CRITICAL_SECTION(op) {
+#define Py_END_CRITICAL_SECTION() }
+#endif
+
 /*
  *  Don't use heap-allocated types for Python < 3.11:  the API needed
  *  to find the dynamic module, 'PyType_GetModuleByDef', was added then.
@@ -1197,7 +1205,19 @@ LB_dealloc(LB* self)
 static PyObject*
 LB_changed(LB* self, PyObject* ignored)
 {
-    LB_clear(self);
+    /* Detach the three cache dicts under a critical section so this cannot race a
+     * concurrent lookup()'s field access (#380).  DECREF them AFTER releasing the
+     * section: a dict dealloc can run arbitrary Python (__del__), which must not
+     * execute while the lock is held. */
+    PyObject *cache, *mcache, *scache;
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    cache = self->_cache;   self->_cache = NULL;
+    mcache = self->_mcache; self->_mcache = NULL;
+    scache = self->_scache; self->_scache = NULL;
+    Py_END_CRITICAL_SECTION();
+    Py_XDECREF(cache);
+    Py_XDECREF(mcache);
+    Py_XDECREF(scache);
     Py_INCREF(Py_None);
     return Py_None;
 }
@@ -1216,6 +1236,44 @@ LB_changed(LB* self, PyObject* ignored)
             cache = c
         return cache
 */
+/* Ensure *field is a dict and return a NEW strong reference to it, taking a critical
+ * section on 'self' so the field's create/read cannot race a concurrent detach in
+ * LB_changed() on a free-threaded build (#380 Q2/Q3).  NO Python runs inside the
+ * section: the candidate dict is allocated before locking, and only a pointer
+ * publish + Py_NewRef happen while held, so it cannot re-enter or deadlock.  The
+ * returned strong reference keeps the dict alive even if LB_changed() detaches the
+ * field immediately afterward.  Returns NULL (error set) on allocation failure.
+ *
+ * NOTE: this guarantees memory safety, not linearizable invalidation -- a lookup that
+ * retains a now-detached cache may still populate/return from that retired generation.
+ * That is acceptable per the cache's semantics (a stale/duplicate cache entry, never a
+ * use-after-free); strict invalidation would require generation checks with callback
+ * and deadlock implications out of scope for this fix. */
+static PyObject*
+_assure_cache_field(LB* self, PyObject** field)
+{
+    PyObject* candidate = NULL;
+    PyObject* ref = NULL;
+
+    /* Allocate outside the lock; publish only if the field is still empty. */
+    if (*field == NULL) {
+        candidate = PyDict_New();
+        if (candidate == NULL)
+            return NULL;
+    }
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    if (*field == NULL && candidate != NULL) {
+        *field = candidate;    /* field takes ownership */
+        candidate = NULL;
+    }
+    if (*field != NULL)
+        ref = Py_NewRef(*field);
+    Py_END_CRITICAL_SECTION();
+    Py_XDECREF(candidate);     /* lost the publish race, or *field was already set */
+    return ref;
+}
+
+
 /* Return a strong reference to a sub-dict of 'cache' for 'key'.
  * Creates a new empty sub-dict if one doesn't exist yet.
  *
@@ -1255,10 +1313,12 @@ static PyObject*
 _getcache(LB* self, PyObject* provided, PyObject* name)
 {
     PyObject* cache;
+    PyObject* parent = _assure_cache_field(self, &self->_cache);  /* strong ref, under lock */
+    if (parent == NULL)
+        return NULL;
 
-    ASSURE_DICT(self->_cache);
-
-    cache = _subcache(self->_cache, provided);  /* strong ref */
+    cache = _subcache(parent, provided);  /* strong ref; user hash runs OUTSIDE the lock */
+    Py_DECREF(parent);
     if (cache == NULL)
         return NULL;
 
@@ -1614,9 +1674,15 @@ _lookupAll(LB* self, PyObject* required, PyObject* provided)
             return NULL;
     }
 
-    ASSURE_DICT(self->_mcache);
-
-    cache = _subcache(self->_mcache, provided);  /* strong ref */
+    {
+        PyObject* parent = _assure_cache_field(self, &self->_mcache);  /* strong ref, under lock */
+        if (parent == NULL) {
+            Py_DECREF(required);
+            return NULL;
+        }
+        cache = _subcache(parent, provided);  /* strong ref; user hash OUTSIDE the lock */
+        Py_DECREF(parent);
+    }
     if (cache == NULL) {
         Py_DECREF(required);
         return NULL;
@@ -1699,9 +1765,15 @@ _subscriptions(LB* self, PyObject* required, PyObject* provided)
             return NULL;
     }
 
-    ASSURE_DICT(self->_scache);
-
-    cache = _subcache(self->_scache, provided);  /* strong ref */
+    {
+        PyObject* parent = _assure_cache_field(self, &self->_scache);  /* strong ref, under lock */
+        if (parent == NULL) {
+            Py_DECREF(required);
+            return NULL;
+        }
+        cache = _subcache(parent, provided);  /* strong ref; user hash OUTSIDE the lock */
+        Py_DECREF(parent);
+    }
     if (cache == NULL) {
         Py_DECREF(required);
         return NULL;
--- a/src/zope/interface/_zope_interface_coptimizations.c
+++ b/src/zope/interface/_zope_interface_coptimizations.c
@@ -1239,10 +1239,11 @@ LB_changed(LB* self, PyObject* ignored)
 /* Ensure *field is a dict and return a NEW strong reference to it, taking a critical
  * section on 'self' so the field's create/read cannot race a concurrent detach in
  * LB_changed() on a free-threaded build (#380 Q2/Q3).  NO Python runs inside the
- * section: the candidate dict is allocated before locking, and only a pointer
- * publish + Py_NewRef happen while held, so it cannot re-enter or deadlock.  The
- * returned strong reference keeps the dict alive even if LB_changed() detaches the
- * field immediately afterward.  Returns NULL (error set) on allocation failure.
+ * section: Py_NewRef and pointer publication cannot re-enter.  Dict allocation is
+ * performed outside the section, after a locked miss, then the field is checked
+ * again before publication.  The returned strong reference keeps the dict alive
+ * if LB_changed() detaches the field immediately afterward.  Returns NULL (error
+ * set) only on allocation failure.
  *
  * NOTE: this guarantees memory safety, not linearizable invalidation -- a lookup that
  * retains a now-detached cache may still populate/return from that retired generation.
@@ -1252,24 +1253,31 @@ LB_changed(LB* self, PyObject* ignored)
 static PyObject*
 _assure_cache_field(LB* self, PyObject** field)
 {
-    PyObject* candidate = NULL;
+    PyObject* candidate;
     PyObject* ref = NULL;
 
-    /* Allocate outside the lock; publish only if the field is still empty. */
-    if (*field == NULL) {
-        candidate = PyDict_New();
-        if (candidate == NULL)
-            return NULL;
-    }
+    /* The first field read must be locked too: LB_changed() writes this pointer. */
     Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
-    if (*field == NULL && candidate != NULL) {
-        *field = candidate;    /* field takes ownership */
-        candidate = NULL;
-    }
     if (*field != NULL)
         ref = Py_NewRef(*field);
     Py_END_CRITICAL_SECTION();
-    Py_XDECREF(candidate);     /* lost the publish race, or *field was already set */
+    if (ref != NULL)
+        return ref;
+
+    /* Allocate outside the lock, then recheck because another lookup may have
+     * published a dict while allocation was in progress. */
+    candidate = PyDict_New();
+    if (candidate == NULL)
+        return NULL;
+
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    if (*field == NULL) {
+        *field = candidate;    /* field takes ownership */
+        candidate = NULL;
+    }
+    ref = Py_NewRef(*field);
+    Py_END_CRITICAL_SECTION();
+    Py_XDECREF(candidate);     /* lost the publish race */
     return ref;
 }
 
--- a/src/zope/interface/tests/test_lookup_concurrency.py
+++ b/src/zope/interface/tests/test_lookup_concurrency.py
@@ -1,0 +1,84 @@
+##############################################################################
+#
+# Copyright (c) 2026 Zope Foundation and Contributors.
+# All Rights Reserved.
+#
+# This software is subject to the provisions of the Zope Public License,
+# Version 2.1 (ZPL).  A copy of the ZPL should accompany this distribution.
+# THIS SOFTWARE IS PROVIDED "AS IS" AND ANY AND ALL EXPRESS OR IMPLIED
+# WARRANTIES ARE DISCLAIMED, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY, TITLE, AND FITNESS FOR A PARTICULAR
+# PURPOSE.
+#
+##############################################################################
+"""Concurrency regression test for the LookupBase cache (see issue #380).
+
+On a free-threaded build, concurrent ``lookup()`` and ``changed()`` calls used to crash the
+interpreter (a data race on the ``_cache``/``_mcache``/``_scache`` fields). This runs the
+reproducer in a subprocess so a regression is observed as a non-zero exit rather than taking
+the whole test run down; on GIL builds it trivially passes.
+"""
+import subprocess
+import sys
+import textwrap
+import unittest
+
+
+_CHILD = textwrap.dedent(
+    """
+    import threading, time
+    from zope.interface import Interface
+    from zope.interface.adapter import AdapterRegistry, AdapterLookup
+
+    class IA(Interface): pass
+    class IB(Interface): pass
+
+    reg = AdapterRegistry()
+    reg.register([IA], IB, "", "v")
+    lk = AdapterLookup(reg)
+    stop = threading.Event()
+
+    def reader():
+        while not stop.is_set():
+            try:
+                lk.lookup([IA], IB, "")
+            except Exception:
+                pass
+
+    def clearer():
+        while not stop.is_set():
+            try:
+                lk.changed(None)
+            except Exception:
+                pass
+
+    threads = ([threading.Thread(target=reader) for _ in range(8)]
+               + [threading.Thread(target=clearer) for _ in range(4)])
+    for t in threads:
+        t.start()
+    time.sleep(3)
+    stop.set()
+    for t in threads:
+        t.join()
+    print("ok")
+    """
+)
+
+
+class ConcurrentLookupChangedTests(unittest.TestCase):
+
+    def test_concurrent_lookup_and_changed_does_not_crash(self):
+        result = subprocess.run(
+            [sys.executable, "-c", _CHILD],
+            capture_output=True, text=True, timeout=60,
+        )
+        self.assertEqual(
+            result.returncode, 0,
+            "concurrent lookup()/changed() crashed the interpreter "
+            "(returncode %r):\n%s" % (result.returncode, result.stderr[-2000:]),
+        )
+        self.assertIn("ok", result.stdout)
+
+
+def test_suite():
+    return unittest.defaultTestLoader.loadTestsFromName(__name__)
--- a/src/zope/interface/tests/test_lookup_concurrency.py
+++ b/src/zope/interface/tests/test_lookup_concurrency.py
@@ -13,10 +13,9 @@
 ##############################################################################
 """Concurrency regression test for the LookupBase cache (see issue #380).
 
-On a free-threaded build, concurrent ``lookup()`` and ``changed()`` calls used to crash the
-interpreter (a data race on the ``_cache``/``_mcache``/``_scache`` fields). This runs the
-reproducer in a subprocess so a regression is observed as a non-zero exit rather than taking
-the whole test run down; on GIL builds it trivially passes.
+On a free-threaded build, concurrent lookup and invalidation calls used to
+crash the interpreter.  Run the reproducer in a subprocess so a regression is
+observed as a non-zero exit rather than taking the whole test run down.
 """
 import subprocess
 import sys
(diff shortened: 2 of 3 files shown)


dataflake:
Please don't forget to sign the contributor agreement, you can do that electronically by following the link Contributor Agreement Form for Zope on the [Plone Foundation contributor agreement page](https://plone.org/foundation/contributors-agreement).

espressolee:
I signed the Zope contributor agreement on 2026-08-13, electronically via sign.plone.org (legal name JUNHYUNG LEE, GitHub @espressolee). Happy to re-send the confirmation if that helps.

dataflake:
Thanks for confirming, I'll check with the team that handles those agreements so that your membership status in the organization reflects it.

icemac (commented):
Reviewed the branch at 5a2fdda: read the full pre/post source and syntax-compiled `_zope_interface_coptimizations.c` against both 3.13 headers (real `Py_BEGIN_CRITICAL_SECTION`) and 3.11 headers (the new fallback macro). Both compile clean apart from the pre-existing `-Wcast-function-type-mismatch` warnings. Refcounting in `_assure_cache_field`, `_getcache`, `_lookupAll` and `_subscriptions` is balanced on every path, including the lost-publish race and the `PyDict_New` failure path, and no Python code runs inside a critical section, so no deadlock or re-entrancy hazard is introduced. The `#ifndef` capability guard and the brace-matching fallback are both right.

The mechanism is sound. My concern is that it is applied to only one of the two `changed()` implementations.

### 1. `VerifyingBase` is left unguarded — blocking

`LB_changed` is now safe, but `verify_changed` (registered as `changed` in `VB_methods`) calls `VB_clear(self)` at line 1976, which calls `LB_clear((LB*)self)` at line 1929, which does bare `Py_CLEAR(self->_cache/_mcache/_scache)` with no critical section.

`VerifyingAdapterLookup` inherits from `VerifyingBase`, and `VerifyingAdapterRegistry` is documented in `adapter.py:981` as "The most commonly-used adapter registry." So: thread A calls `VerifyingAdapterLookup.lookupAll()` → `_assure_cache_field` locks `self` and reads/publishes `self->_mcache`; thread B calls `.changed(None)` → `verify_changed` → `VB_clear` → unlocked `Py_CLEAR(self->_mcache)`. That is exactly the race from #380, unmitigated — same `PyMutex_Unlock: unlocking mutex that is not locked` / segfault.

`verify_changed` needs the same detach-under-lock, DECREF-after-unlock treatment as `LB_changed`. Note it can't simply reuse `VB_clear`, since `VB_clear` is also `tp_clear`.

### 2. `_verify()` reads `_verify_ro` unsynchronized against `verify_changed`

`_verify` NULL-checks `_verify_ro` and `_verify_generations` (line 2020), then does `PyTuple_GET_SIZE(self->_verify_ro)` (2022) and `PyTuple_GET_ITEM(self->_verify_ro, i)` (2031) on the raw field with no lock and no strong reference, and `PyObject_GetAttr` inside the loop can run arbitrary Python.

If thread B runs `verify_changed` → `VB_clear` → `Py_CLEAR(self->_verify_ro)` between the NULL check and `PyTuple_GET_SIZE`, or between iterations, thread A dereferences NULL or indexes a tuple whose last reference has just been dropped. `_verify` is reached by every `VerifyingAdapterLookup.lookup()`, and it calls `changed()` itself on generation mismatch, so read-only concurrent traffic is enough to trigger it. Same bug class as #380, and arguably part of the same fix.

Smaller points are inline.

— _Comment created by Claude_

src/zope/interface/tests/test_lookup_concurrency.py
```diff
@@ -0,0 +1,136 @@
...
+
+On a free-threaded build, concurrent lookup and invalidation calls used to
+crash the interpreter.  Run the reproducer in a subprocess so a regression is
+observed as a non-zero exit rather than taking the whole test run down.
+"""
+import subprocess
+import sys
+import textwrap
+import unittest
+
```
icemac, line 25: The regression test can't catch the two issues from the review summary. The child script subclasses plain `LookupBase` and calls `lk.changed(None)`, which routes to the now-guarded `LB_changed`; it never exercises `VerifyingBase`/`verify_changed`, i.e. the path that is still broken and the one most users actually hit. A second worker set built on `VerifyingAdapterLookup` (or on `VerifyingBase` with a stub `_registry`) would make the suite cover the reported crash surface.

Minor, while you're here: the child hard-sleeps `time.sleep(3)` with 13 spinning threads, adding 3+ s to every free-threaded CI run, and it keeps sleeping the full 3 s even after `record_error()` has set `stop`. `stop.wait(3)` is equivalent on success and returns immediately on failure.

— _Comment created by Claude_

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -1216,6 +1236,52 @@ LB_changed(LB* self, PyObject* ignored)
...
+ * if LB_changed() detaches the field immediately afterward.  Returns NULL (error
+ * set) only on allocation failure.
+ *
+ * NOTE: this guarantees memory safety, not linearizable invalidation -- a lookup that
+ * retains a now-detached cache may still populate/return from that retired generation.
+ * That is acceptable per the cache's semantics (a stale/duplicate cache entry, never a
+ * use-after-free); strict invalidation would require generation checks with callback
+ * and deadlock implications out of scope for this fix. */
+static PyObject*
+_assure_cache_field(LB* self, PyObject** field)
```
icemac, line 1235: `_assure_cache_field` takes the object's critical section unconditionally, even to read an already-populated `*field`. On a free-threaded build that means all `lookup`/`lookup1`/`adapter_hook`/`queryAdapter`/`lookupAll`/`subscriptions` calls against one registry serialize on a single object mutex, in steady state with a warm cache.

Adapter lookup is the hottest path in `zope.component`, so this trades the crash for a scalability ceiling on exactly the workload free-threading is meant to help. An atomic acquire load plus `PyUnstable_TryIncRef`, falling back to the locked path on failure, would keep the hot path lock-free. At minimum it seems worth a benchmark before merging.

— _Comment created by Claude_

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -1216,6 +1236,52 @@ LB_changed(LB* self, PyObject* ignored)
             cache = c
         return cache
 */
+/* Ensure *field is a dict and return a NEW strong reference to it, taking a critical
+ * section on 'self' so the field's create/read cannot race a concurrent detach in
+ * LB_changed() on a free-threaded build (#380 Q2/Q3).  NO Python runs inside the
```
icemac: `(#380 Q2/Q3)` doesn't refer to anything in #380 — looks like a leftover from drafting, and it will send anyone tracing the rationale on a detour.

Also, this 12-line block comment is longer than the function it documents. The "memory safety, not linearizable invalidation" paragraph is the genuinely valuable part; the rest might read better in `CHANGES.rst` or on the issue.

— _Comment created by Claude_

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -71,6 +71,14 @@ _PyDict_GetItemRef(PyObject *p, PyObject *key, PyObject **result)
             return NULL;                                                       \
     }
```
icemac: `ASSURE_DICT` is dead code now — its three call sites were the ones replaced by `_assure_cache_field`, and nothing else references it. Worth deleting as part of this change, otherwise it stays available for a future contributor adding a fourth cache field to reach for, silently reintroducing the #380 pattern.

— _Comment created by Claude_

espressolee:
Thank you for the careful pass. Both blocking paths are real at `5a2fdda`, and
I pushed `5ec2eda9f9e93f8a2c09485684eb1344efcccfa9` to address them.

The updated regression module now includes `VerifyingBase` cache workers and a
deterministic `_verify()` snapshot test. Against exact predecessor `5a2fdda`,
the cache worker terminates with SIGSEGV and the snapshot test raises
`SystemError: bad argument to internal function`. At the new head, the two-test
module passes in 10/10 fresh processes on CPython 3.14.6t with the GIL disabled.

Implementation changes:

- `verify_changed()` detaches all three cache fields and both verification
  fields under the object critical section, then decrefs after unlocking.
- The newly computed `_verify_ro` / `_verify_generations` pair is published
  together under the same lock.
- `_verify()` takes strong references to one pair while locked and uses only
  that pair after unlocking, including across `PyObject_GetAttr()` and rich
  comparison callbacks.
- `LB_clear()` / `VB_clear()` are left for GC and deallocation only.
- I removed dead `ASSURE_DICT`, removed the stale `#380 Q2/Q3` shorthand,
  shortened the helper comment, and changed the child wait to `stop.wait(3)`.
- While touching `_generations_tuple()`, I also added its missing
  `PyTuple_New()` failure check.

Validation at the new head:

- CPython 3.14.6t C suite: 1,369 tests, 0 failures, 7 skipped.
- CPython 3.12.11 C and pure suites: 1,373 tests each, 0 failures, 9 skipped.
- Changed-file pre-commit checks pass.
- The C file syntax-compiles against CPython 3.13 and 3.11 headers.

I also measured the warm-cache concern rather than assuming the lock cost was
negligible. These are bounded local microbenchmarks on macOS arm64 / CPython
3.14.0rc1t, five alternating fresh-process samples per revision:

- one thread, one warm `lookup1` cache: candidate/base median ratio `0.900`;
- three threads split across `_cache`, `_mcache`, and `_scache`: ratio `0.984`;
- eight threads sharing one warm cache: candidate `4.44M` versus unsafe base
  `0.98M` calls/s. The candidate itself did not scale above its one-thread
  throughput, while the unsafe base degraded much more.

So the measurement shows a bounded single-thread cost and a same-object
ceiling, but it does not show worse contended throughput than the current code;
there is no safe lock-free comparator here. I did not adopt
`PyUnstable_TryIncRef`: it is available only from 3.14, and it does not by
itself make an unsynchronized field-pointer load/publication safe on the 3.13+
support range. I have kept the simple public critical-section design and leave
the performance trade-off visible for maintainer judgment.

Could you please take another look at the corrected lifetime paths?

icemac (commented):
Reviewed the whole diff for correctness. **No correctness bug found in the fix** — refcount ownership is balanced on all success and error paths, no allocation/callback/re-entrant decref happens inside a critical section, and there is no same-object nesting (so no deadlock path). I worked through the interleavings of a lookup racing an invalidation and could not construct a stale read: taking a strong ref to the parent dict *before* computing uncached results means the racing lookup writes into the detached dict instead.

Checks run:

- Full suite on CPython 3.14 (GIL) and free-threaded 3.15.0rc1t: 1133 tests, OK on both.
- The new tests applied to a `master` checkout on 3.15t: **both fail** (`SystemError: bad argument to internal function` for the snapshot test), so they are genuine regressions.
- Compiled `-Wall -Wextra` against 3.10–3.15 headers, including a forced `#ifndef Py_BEGIN_CRITICAL_SECTION` fallback build: no new diagnostics.
- All three `ASSURE_DICT` call sites converted and the macro fully removed.

Three low-severity notes inline — none blocking.

One out-of-scope observation: a multithreaded stress run over `AdapterRegistry`/`VerifyingAdapterRegistry` on 3.15t showed no C-level crashes on this branch, but did hit pre-existing *Python-level* races (`RuntimeError: dictionary changed size during iteration` in `adapter.py`, `KeyError` in `interface.py`). Not this PR's problem, but they suggest `Py_MOD_GIL_NOT_USED` remains optimistic for the package as a whole.

— _Comment created by Claude_

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -1914,16 +1995,24 @@ verify_changed(VB* self, PyObject* ignored)
...
 
+    /* Another changed() may have published a newer pair while Python ran
+     * above.  Replace both fields atomically and retire that pair afterward. */
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    replaced_ro = self->_verify_ro;
+    replaced_generations = self->_verify_generations;
     self->_verify_ro = ro;
+    self->_verify_generations = generations;
+    Py_END_CRITICAL_SECTION();
```
icemac, line 2011: This publish is last-writer-wins, so an *older* snapshot can overwrite a newer one.

Thread A enters `changed()` at generation 5, clears the caches, builds `S5`, then stalls. The registry advances to 6; thread B's `changed()` publishes `S6` and lookups repopulate the caches with generation-6 data. Thread A then overwrites the pair with `S5`.

The effect is bounded and self-healing — the next `_verify()` sees current(6) != stored(5), calls `changed()` again and flushes, so it costs one redundant invalidation rather than serving stale data. But the comment ("Another `changed()` may have published a newer pair … Replace both fields") reads as if clobbering the newer pair were the intent. Worth either saying explicitly that the redundant invalidation is acceptable, or discarding the older result (e.g. compare against a monotonic publish counter captured before the Python calls).

— _Comment created by Claude_

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -1216,6 +1227,42 @@ LB_changed(LB* self, PyObject* ignored)
...
+_assure_cache_field(LB* self, PyObject** field)
+{
+    PyObject* candidate;
+    PyObject* ref = NULL;
+
+    /* The first field read must be locked too: LB_changed() writes this pointer. */
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    if (*field != NULL)
+        ref = Py_NewRef(*field);
+    Py_END_CRITICAL_SECTION();
```
icemac, line 1244: The fast path takes the lookup object's `ob_mutex` just to read an already-non-`NULL` pointer and incref it, which puts a lock acquisition on every `lookup()`/`lookupAll()`/`subscriptions()`/`queryAdapter()` call.

Since one `VerifyingAdapterLookup` is shared process-wide, on a free-threaded build this turns a previously lock-free cache read into a single serialization point for exactly the workload the cache exists to parallelize. GIL builds are unaffected (the macro is a no-op) and the section is tiny, so the impact is bounded — but it seems worth measuring on 3.14t before release.

If it does matter, two options: an atomic load plus try-incref on the fast path, or have `changed()` call `PyDict_Clear()` on the existing dicts instead of detaching them. The latter also matches the `self._cache.clear()` semantics of the pure-Python `LookupBase` and would stop in-flight lookups from writing results into an orphaned dict.

— _Comment created by Claude_

src/zope/interface/_zope_interface_coptimizations.c
```diff
@@ -1891,9 +1955,26 @@ _generations_tuple(PyObject* ro)
...
+    PyObject *old_ro, *old_generations;
+    PyObject *replaced_ro, *replaced_generations;
+    PyObject *t, *ro, *generations;
+
+    /* Clear all lookup and verification fields as one state transition.  The
+     * detached objects remain alive until after the critical section. */
+    Py_BEGIN_CRITICAL_SECTION(OBJECT(self));
+    cache = self->lookup._cache;   self->lookup._cache = NULL;
+    mcache = self->lookup._mcache; self->lookup._mcache = NULL;
+    scache = self->lookup._scache; self->lookup._scache = NULL;
```
icemac, line 1968: This hand-detaches the three `LB` cache fields rather than reusing `LB_changed()`. Correct today, but the fields are now enumerated in four places: `LB_traverse`, `LB_clear`, `LB_changed` and here.

Adding a fourth cache field to `LB` later would compile fine while silently leaving it un-invalidated on every `VerifyingBase.changed()` — the same class of bug this PR fixes. The Python implementation avoids this by chaining to `LookupBaseFallback.changed(self, ...)`; calling `LB_changed((LB*)self, ignored)` first and keeping only the `_verify_*` pair swap here would too.

— _Comment created by Claude_

icemac:
@tseaver Could you please have a final look into the changes? You understand the C code way better than me.

src/zope/interface/_zope_interface_coptimizations.c
espressolee: Measured — preliminary medians on macOS arm64 (M4 Pro, unpinned), so not portable to x86; pinned x86 numbers to follow. CPython 3.14.0rc1t, `5ec2eda9` vs `2bd6aec4`, 6 alternating fresh processes × 3 repeats per cell, warm `lookup()` on one shared `VerifyingAdapterRegistry`. Head 3.86M/3.19M/3.22M calls/s at 1/2/4 threads; base (no lock) 4.36M/2.77M/2.37M. Spreads are wide (~2× in some cells), so the safe reading is: head does not scale, base scales worse, and head is not slower than base under contention on this host. Single-thread cost ~5% plain, ~12–17% verifying, the latter including the `_verify()` snapshot section.

`PyDict_Clear()` alone would not remove the lock: results land in the per-`provided` subcache (opt.c:1408/1698/1789) and the fields start `NULL`, so a lock-free fast path needs a publication protocol, not just a load. Not changing the fast path in this PR.

src/zope/interface/_zope_interface_coptimizations.c
espressolee: Adopted: `verify_changed()` now calls `LB_changed((LB*)self, ignored)` first and keeps only the `_verify_*` detach and republish, so the cache fields are enumerated in `LB_traverse`, `LB_clear` and `LB_changed` only — pushed as `37f33172245732048c7e3236dc097a3e3bda4790` on top of `5ec2eda9`.

The comment records the consequence rather than claiming equivalence: invalidation is now two transitions, so a concurrent reader can see cleared caches while the previous pair is still installed, and verifies against that pair as before. Bounded, not proven equal; flagged for the C pass.

Suite with the change, macOS arm64: 3.14.6t GIL-disabled and 3.12.11 C/pure all pass, `test_lookup_concurrency` 10/10 fresh processes; `clang -Wall -Wextra` unchanged at the 12 pre-existing warnings.

src/zope/interface/_zope_interface_coptimizations.c
espressolee: Took the comment option — comment-only edit at `_zope_interface_coptimizations.c:2004` in the same push. The comment now names last-writer-wins and states the redundant invalidation as the accepted cost.

A publish counter would not buy freshness: `_generations_tuple()` runs Python before the store, so the winning pair's generations can already trail the registry, and `_verify()` catches that the same way it catches an older pair — by generation mismatch on the next call.

jensens:
Additional review pass on c82b81d. Existing threads read; overlaps marked ↔.
Scope: changes in this PR only. Pre-existing races in `VerifyingBase.changed()` found along the way → #384.

### Summary

- Crash fix: no objection. New tests pass on 3.14.2t, skip on 3.14.3 (GIL).
- 1: two new code comments describe pre-existing races as benign → reword.
- 2–5: test robustness.
- 6: hot-path data point.
- 7–10: cleanup.
- Nothing blocking from my side.

### Code comments

**1. Two new comments describe pre-existing races as benign** — [L1963–L1969][c1963], [L2008–L2012][c2008]

- [L2008–L2012][c2008] ("the cost is a redundant invalidation"): holds only if both pairs share the same `ro`. Different `ro` (rebase racing another `changed()`) → `_verify()` polls the wrong registries → never a mismatch → stale results stay. ↔ L2011 thread.
- [L1963–L1969][c1963] ("verifies against that pair as before"): a reader can still store a result computed against older base-registry state into the new live cache; the stored generations then match → no invalidation. ↔ L1968 thread ("bounded, not proven equal; flagged for the C pass"), ↔ review of 2026-09-08 ("could not construct a stale read").
- Both reproduce on unmodified master (dd65a10), also on a GIL build → not caused by this PR, out of scope here. Repro scripts and fix sketch: #384.
- Here: reword both comments to "known limitation, see #384".

### Tests

**2. `test_verify_uses_one_strong_snapshot` is skipped on GIL builds without need** — [L247][t247], child gate [L160][t160]

- Test is deterministic, independent of the GIL (`Event.wait()` releases it inside the `_generation` property).
- With the gate removed on 3.14.3 (GIL): master C source → `SystemError: Objects/object.c:1101: bad argument to internal function`; PR → `ok`.
- → the `_verify()` snapshot fix has no regression coverage on the 3.10–3.15 GIL matrix.
- Proposal: gate this test on the C implementation only; keep the GIL gate for the stress test.
- CHANGES.rst says "free-threaded (no-GIL) data races"; the `_verify()` part also affected GIL builds.

**3. Child process does not inherit the parent's `sys.path`** — [L256][t256]

- Parent found `zope.interface` via in-process path injection (buildout / `zc.recipe.testrunner` `bin/test`) → child: `ModuleNotFoundError: No module named 'zope'` → both tests fail instead of pass/skip.
- Fix: `env={**os.environ, "PYTHONPATH": os.pathsep.join(sys.path)}`.

**4. Deadlock regression surfaces only as an opaque `TimeoutExpired`** — [L220][t220], [L145][t145]

- Worker threads are non-daemon; `_CACHE_CHILD` uses `t.join()` without timeout.
- Snapshot child: `thread.join(10)` → `SystemExit("verification reader did not finish")` → interpreter shutdown joins the blocked non-daemon thread → hang → parent `timeout=60` → `TimeoutExpired`; label and stderr never reach the assertion message.
- Fix: `daemon=True` + bounded joins, or `os._exit()`.

**5. Crash regression is a fixed 3 s probabilistic stress run** — [L142][t142]

- ↔ `time.sleep(3)` thread (`stop.wait(3)` adopted): remaining point is reliability, not duration.
- 16 threads on a 2–4 core runner may never hit the detach/incref window → green on broken code.
- Deterministic alternative: cached value whose `__del__` blocks → pauses `LB_changed` between detach and decref (same technique as the snapshot test).

### Performance

**6. Hot-path cost** — [L2041][c2041], [L1241][c1241]

- ↔ L1244 thread (already measured by the author on arm64: ~5 % plain, ~12–17 % verifying).
- x86_64 data point, 3.14.2t, single thread, cached `lookup()`, PR vs. master C source: `LookupBase` 7.40 → 6.38 M/s (−14 %), `VerifyingBase` 5.23 → 4.07 M/s (−22 %).
- Cached `VerifyingBase` lookup: 2 lock acquisitions, 3 atomic incref/decref pairs on shared objects. Fetching pair + cache in one critical section would halve that → fits the fix in #384.

### Cleanup

**7. Silent `#ifndef` fallback** — [L70–L73][c70]

- If `Py_BEGIN_CRITICAL_SECTION` is ever not visible on a free-threaded build (e.g. future limited-API / abi3t build), the critical sections become plain blocks and the #380 crash returns without a compile-time signal.
- Fix: `#ifdef Py_GIL_DISABLED` → `#error` inside the fallback.

**8. Duplicated parent-ref block** — `_lookupAll` [L1667][c1667], `_subscriptions` [L1758][c1758], variant in `_getcache` [L1305][c1305]

- Fetch parent → `NULL` check → `_subcache()` → decref parent, three times by hand.
- Helper, e.g. `_field_subcache(self, &self->_mcache, provided)`, returning the sub-dict.

**9. `verify_changed()` calls the Python-facing `LB_changed(self, ignored)`** — [L1970–L1973][c1970]

- ↔ L1968 thread (reuse adopted on request; single enumeration of cache fields is right).
- Refinement: `LB_changed()` cannot return `NULL` → the check at L1971 is dead code; each call increfs/decrefs `None`.
- Static helper that detaches the three fields into out-params (caller holds the lock, decrefs after), used by both `LB_changed` and `verify_changed`. Keeps the single enumeration.

**10. `test_suite()` is dead code** — [L279][t279]

- CI and tox run `python -m unittest discover -s src/zope`; no other module in `src/zope/interface/tests` defines `test_suite()`.
- Never executed → uncovered lines against `fail_under = 98`. On GIL builds the body of `_run_child` after `skipTest` is uncovered as well.

---

- Method: automated `/code-review` pass in a sub-agent (diff + full C source at c82b81d), then a second verification pass in the main session, including reading the existing review threads on this PR.
- Environment: Linux x86_64, Clang 21.1.4, CPython 3.14.2t (free-threaded) and 3.14.3 (GIL). Baseline: master (dd65a10).
- Executed: 2–3 reproduced with scripts in the first pass (not re-run in the second). Races behind 1: deterministic scripts, re-run in the second pass, see #384. New test module run on both interpreters.
- Measured: 6 from a micro-benchmark, single 1.5 s run per cell, unpinned laptop CPU (i7-11370H) → indicative only.
- Code reading only, not executed: 4, 5, 7–10.
- Proposed fixes are neither implemented nor tested. Full test suite not run.

[c70]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L70-L73
[c1241]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L1241
[c1305]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L1305
[c1667]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L1667
[c1758]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L1758
[c1963]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L1963-L1969
[c1970]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L1970-L1973
[c2008]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L2008-L2012
[c2041]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/_zope_interface_coptimizations.c#L2041
[t142]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L142
[t145]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L145
[t160]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L160
[t220]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L220
[t247]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L247
[t256]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L256
[t279]: https://github.com/zopefoundation/zope.interface/blob/c82b81d1298304f3d8995647212b212a3c8d695c/src/zope/interface/tests/test_lookup_concurrency.py#L279

espressolee:
Thanks for the detailed review. I reproduced the points that would change code before changing any, and pushed `497b541`, `890d808`, `d385b9e`. Measurements and remaining limitations below.

All local measurements are macOS aarch64; the CI results are reported separately at the end. I did not test Linux, Windows, PyPy, CPython 3.11–3.13, or a two-core runner locally.

## 2 — confirmed

The child extracted verbatim with only the GIL term removed, 5 runs per cell, the column being the success count:

| GIL interpreter | master `dd65a10` | this branch |
|---|---|---|
| 3.10.20 | 0/5 ok — `SystemError: Objects/object.c:736` | 5/5 ok |
| 3.14.6 | 0/5 ok — `Objects/object.c:1104` | 5/5 ok |
| 3.15.0b4 | 0/5 ok — `Objects/object.c:1106` | 5/5 ok |

It distinguishes master from this branch on all three locally tested GIL versions, so it is not a new-CPython artifact and gating on the C implementation adds coverage beyond 3.14+. Two notes for anyone re-running: the `object.c` locations differ across CPython versions, so your `:1101` on 3.14.3 and the three above are the same error at different lines; and the child's own guard needed the same term removed, since `sys._is_gil_enabled` does not exist before 3.13.

`CHANGES.rst` is split in two accordingly. You are right that describing the whole change as "free-threaded (no-GIL) data races" was understating it.

## 3 — confirmed, and narrowed

Reproduced by mechanism rather than by running buildout: a clean venv without `zope.interface` installed, plus an in-process `sys.path.insert` of `src/`. Before, `run=2 err=0 fail=2 skip=0`, both `ModuleNotFoundError: No module named 'zope'`. After, `run=2 err=0 fail=0 skip=0` and the reproducers actually execute. Not skips — failures, as you said.

CI is not affected: the workflow installs editable and runs `coverage run -p -m unittest discover -s src/zope`. `buildout.cfg` does declare `zc.recipe.testrunner`, so the local `bin/test` path is where this bites. Fixed as you proposed, with a comment recording that `PYTHONPATH` is a workaround and not an exact copy of the parent's import state — it cannot carry an entry containing `os.pathsep` and it places those paths ahead of the child's standard library.

## 5 — I did not observe a false pass, but I cannot reach your runner

I observed no false passes on free-threaded CPython 3.14.6, including under macOS background QoS, which confines the process to this machine's four efficiency cores. **That does not establish how the test behaves on a two-core Linux runner**, which is what you were asking about.

`_CACHE_CHILD` verbatim against the broken subject, `master` @ `dd65a10`:

| batch | outcome on 20 (or 12) runs |
|---|---|
| unrestricted, 12 cores | 20 crashed, 0 passed |
| efficiency cores only, first batch | 18 crashed, **2 hit my own 90 s cap** — not observed crashes |
| efficiency cores only, further probe with a 300 s cap | 12 crashed, 0 passed; 2.7 s to 6.0 s, none over 60 s |
| efficiency cores only, second batch | 20 crashed, 0 passed |
| **total** | **72 runs, 0 passed; 70 crashed, 2 hit the cap** |

This branch was 0 crashes in 60 runs across the same batches.

Your *proposal* is the part I would keep regardless, and I have not implemented it: a deterministic test built on a blocking `__del__` would address the scheduling dependence directly rather than relying on a window being hit.

## 4 — I could not produce it

The two runs that hit my 90 s cap above are the only hang-shaped outcome I saw, and they did not recur in either later batch. I cannot determine the cause from these measurements, and I have not established that they reproduce the hang you described.

So I have no evidence for it. Your argument from the source — non-daemon workers plus `t.join()` with no timeout — is untouched by that; I could not produce the failure and did not want to claim I had. I did take the smaller half of the same concern: `resume.set()` now runs in a `finally`, so a `changed()` that raises is reported as itself rather than as the reader's `resume.wait(10)` expiring.

## 1, 7, 9, 10 — taken

**1.** Both comments reworded to state the limitation and reference #384 instead of calling it benign. I have not reproduced those races myself; the rewording reflects your measurement, and the comments say so.

**7.** Now a hard error, and here is what the guard is and is not worth. Forcing the fallback on at `c82b81d1` and rebuilding free-threaded gave 20 crashes in 20 lookup/changed runs (16 SIGSEGV, 4 SIGABRT) against 0 in 20 unforced, so the hazard is immediate rather than hypothetical. But the snapshot reproducer passed 20/20 **with the fallback forced**, so that probe did not expose a snapshot failure and the guard is protecting the cache-lifetime half specifically. It also cannot fire where `Py_GIL_DISABLED` is absent from the extension's translation unit, so it does not detect every such build.

**9.** Taken, via the helper you suggested. `LB_detach_caches()` is now used by both `LB_changed()` and `verify_changed()`, which keeps the single enumeration and drops both the unreachable NULL check and the `None` round trip. The NULL check at the other `changed()` call site stays — that one goes through `PyObject_CallMethodObjArgs()` and can genuinely fail.

**10.** Dropped.

## 6 and 8 — not in this revision

**6.** I have not re-measured your x86_64 numbers and am not disputing them; you flagged them as indicative and mine were arm64. I have deferred folding the pair fetch and the cache into one critical section because #384 may change the same critical-section boundaries; I can include it here if you prefer.

**8.** Left out to keep this revision focused on correctness, but I can include it if you prefer.

## State

The single CI run on `d385b9e` reports 67 successful checks, 1 failure and 2 skipped, of 70. The CPython jobs passed across the configured platforms — `3.10` through `3.15` on `ubuntu-latest`, `windows-latest`, `windows-11-arm` and `macos-latest`, `3.14t` and `3.15t` on three platforms, and `pypy-3.11 / ubuntu-latest`. That is cross-platform evidence for the gate change, but one run does not measure flakiness, and passing checks do not by themselves show which tests executed in each job.

Locally: 3.14.6t `Ran 1341, OK`; 3.14.6 GIL `Ran 1341, OK`; the concurrency module `Ran 2, OK`, zero skips free-threaded and one on GIL. On 3.10.20 the full suite reports 6 errors, but the identical set and count appear at `c82b81d1` on the same interpreter, so these edits did not introduce that error set relative to the previous revision.

## The PyPy Windows installation failure also occurs on master

I suggested earlier that `test (pypy-3.11, windows-latest)` just needed a re-run. That was too confident. The same signature appears three times:

```
ERROR: Could not install packages due to an OSError:
('Connection broken: IncompleteRead(929479 bytes read, 13018 more expected)', ...)
```

on this PR on 2026-09-16 and again today, and on **`master` at `dd65a10` on 2026-09-20**, with none of these changes present. In the current run it occurs during `Install zope.interface`, after `wheel` and `packaging` install from cache and before any test executes; `pypy-3.11 / ubuntu-latest` is unaffected. Its occurrence on master shows these changes are not required to trigger it. I have not identified which package was being downloaded, or the cause.

Given the recurrence on both branches I think a separate issue would be useful, but that is a maintainer call and I have not opened one.

espressolee:
Follow-up on point 5. I said above that I could only approximate your 2–4 core runner with macOS background QoS. I have now tested a closer one: under Docker Desktop on macOS aarch64, `docker run --cpuset-cpus=0,1` restricts the container to two Linux guest CPUs (`visible cpus: 12, affinity: 2`). Still not a particular two-core CI runner's hardware or scheduling, but closer than what I had.

Linux, free-threaded CPython 3.14.7, `_CACHE_CHILD` verbatim, affinity 2:

| subject | 20 / 12 runs | whole-child wall time |
|---|---|---|
| `master` @ `dd65a10` (broken) | **20 crashed**, 0 timed out, 0 exited "ok" | mean 2.5 s, min 1.5 s, **max 3.1 s** |
| this branch (fixed) | 0 crashed, **12 passed** | mean 3.1 s, min 3.1 s, max 3.2 s |

I observed no false greens in this batch. That does not rule them out on your runners — you said "may", and 20 runs do not refute a possibility.

**The second row is the useful one, because it decomposes the first.** A clean run is startup plus `stop.wait(3)` plus teardown and join, and it costs 3.1 s — so everything outside the three-second window is about 0.1 s here. A crash terminates the child, so a crashing run is startup plus time-to-crash. Subtracting:

- fastest catch: 1.5 s → roughly 1.4 s into the window
- **slowest catch: 3.1 s → roughly 3.0 s, the whole window**

So the test is not failing to reach the race at two CPUs — it is reaching it anywhere from halfway through its budget to the very end of it. That is a better argument for your deterministic `__del__` proposal than the catch rate was: the budget is already fully consumed in the worst observed case, and a slower runner has nothing left to spend.

Two notes on method. My first attempt used `docker run --cpus=2`, which sets a CFS quota rather than an affinity; the container still reported affinity 12, and crashed in 20/20 at a wall time rounding to 0.0 s. Quota and affinity constrain different things, and only the second is what you were describing. And the timings are whole-child wall clock, so the decomposition above rests on the fixed branch being a fair overhead baseline for the broken one — same image, same interpreter, same container, same batch.

One batch each, one host, a Linux VM rather than bare metal. I have not implemented the deterministic test.

jensens (approved):
Re-checked d385b9e locally (Linux x86_64, gcc 13, CPython 3.14.2t and 3.14.3):

- Concurrency module: 3.14.2t `Ran 2, OK`; 3.14.3 `Ran 2, OK (skipped=1)`.
- 3: parent with in-process `sys.path.insert(0, "src")`, package not installed → both tests execute on both interpreters (before: 2 failures). Confirmed.
- 7: `#undef Py_BEGIN_CRITICAL_SECTION` ahead of the guard → `#error` with 3.14t headers, clean with 3.14 headers. Confirmed.
- Full suite with the `test` extra: 3.14.2t `Ran 1369, OK (skipped=7)`; 3.14.3 `Ran 1369, OK (skipped=8)`.
- CI: the attempt-1 `pypy-3.11 / windows-latest` failure is the same `Install zope.interface` step that failed on master dd65a10 on 2026-09-20; attempt 2 is green. Prior history: #378, zopefoundation/meta#436. Not this PR.
- #384 reproducers unchanged on d385b9e, as expected; the reworded comments match the code.

- 6 and 8: leave out of this PR. 6 moves the same critical-section boundaries as the #384 fix and belongs there; 8 can go with it.
- 4 and 5: follow-up, not blocking. Your 2-CPU decomposition (worst catch at ~3.0 s of a 3.0 s budget) is the stronger argument for the deterministic `__del__` variant; 4 (daemon threads, bounded joins) fits the same small PR.
- Optional: CHANGES.rst, second entry — the macOS/version list reads like a lab note; "also affects GIL builds" is enough.

Approving.

---
AI-assisted: re-verification with Claude Fable 5.1 via Claude Code, same setup as in my previous comment; read and approved by me before posting.

jensens (approved):
Re-checked d385b9e locally (Linux x86_64, gcc 13, CPython 3.14.2t and 3.14.3):

- Concurrency module: 3.14.2t `Ran 2, OK`; 3.14.3 `Ran 2, OK (skipped=1)`.
- 3: parent with in-process `sys.path.insert(0, "src")`, package not installed → both tests execute on both interpreters (before: 2 failures). Confirmed.
- 7: `#undef Py_BEGIN_CRITICAL_SECTION` ahead of the guard → `#error` with 3.14t headers, clean with 3.14 headers. Confirmed.
- Full suite with the `test` extra: 3.14.2t `Ran 1369, OK (skipped=7)`; 3.14.3 `Ran 1369, OK (skipped=8)`.
- CI: the attempt-1 `pypy-3.11 / windows-latest` failure is the same `Install zope.interface` step that failed on master dd65a10 on 2026-09-20; attempt 2 is green. Prior history: #378, zopefoundation/meta#436. Not this PR.
- #384 reproducers unchanged on d385b9e, as expected; the reworded comments match the code.

- 6 and 8: leave out of this PR. 6 moves the same critical-section boundaries as the #384 fix and belongs there; 8 can go with it.
- 4 and 5: follow-up, not blocking. Your 2-CPU decomposition (worst catch at ~3.0 s of a 3.0 s budget) is the stronger argument for the deterministic `__del__` variant; 4 (daemon threads, bounded joins) fits the same small PR.
- Optional: CHANGES.rst, second entry — the macOS/version list reads like a lab note; "also affects GIL builds" is enough.

Approving.

---
AI-assisted: re-verification with Claude Fable 5.1 via Claude Code, same setup as in my previous comment; read and approved by me before posting.

icemac:
Thank you for this PR and for working on it/reviewing it. 😃

